PR #1, opened 2020-03-16
Alter workspace update script to use new mapping, handle BAIs/MD5s

ThePletch:
* Adds new buckets involved in migration to bucket list
* Alters predicate to identify paths to update to correctly mark BAI and MD5 files for update
* Removes regex parsing logic from lookup tool - new mapping is now a direct path-to-path lookup

### DIFF
--- a/scripts/update_workspace_dd.py
+++ b/scripts/update_workspace_dd.py
@@ -7,8 +7,6 @@ import pandas as pd
 from firecloud import api as fapi
 from datetime import datetime
 import csv
-import sys
-import re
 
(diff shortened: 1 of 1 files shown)


ThePletch:
This removes the code that parses out a sample key for each path it tries to update, which was useful in enumerating the samples we couldn't transfer (but wasn't adequate to identify file extension, for instance). Instead, files we were unable to migrate will be identified by their full source path. Is that acceptable, or are we worrried it will be an overwhelming amount of text for users?

PR #2, opened 2020-03-16
Fix list presence predicate for summary dataframes

ThePletch:


scripts/update_workspace_dd.py
```diff
@@ -4,6 +4,7 @@
 import shutil
 import argparse
 import pandas as pd
+import numpy
```
mmorgantaylor: for convention, suggest
`import numpy as np`

scripts/update_workspace_dd.py
```diff
@@ -381,9 +382,9 @@ def get_replacement_path(original_path, mapping):
 
 def summarize_results(df_paths, do_replacement=True):
     # get some summary stats
-    n_valid_extension_paths_to_replace = len(df_paths[df_paths['file_type'] in EXTENSIONS_TO_MIGRATE])
+    n_valid_extension_paths_to_replace = len(df_paths[df_paths['file_type'].isin(EXTENSIONS_TO_MIGRATE)])
     n_paths_updated = len(df_paths[df_paths['update_status'] == 200])
-    n_nonvalid_extension_paths = len(df_paths[df_paths['file_type'] not in EXTENSIONS_TO_MIGRATE])
+    n_nonvalid_extension_paths = len(df_paths[numpy.logical_not(df_paths['file_type'].isin(EXTENSIONS_TO_MIGRATE))])
```
mmorgantaylor: and then change `numpy` to `np` here

PR #4, opened 2020-06-12
Add support for mapping reduced bam paths to unreduced paths where required

ThePletch:
We've seen cases where users are trying to map paths pointing at the reduced version of a bam which has since been cleaned up. Since we can't build a comprehensive list of all possible bam varieties that might have been in a directory, I instead introduced this system to try mapping other variations of the path if we can't map the path they've given us.

### DIFF
--- a/scripts/update_workspace_dd.py
+++ b/scripts/update_workspace_dd.py
@@ -15,7 +15,10 @@ EXTENSIONS_TO_MIGRATE = ['bam', 'bai', 'md5']
 # and try to find it in the mapping again before giving up (e.g. for files that have been run
 # through specific process, like reduced bams)
 FALLBACK_REPLACEMENTS = {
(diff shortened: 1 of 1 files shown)


scripts/update_workspace_dd.py
```diff
@@ -450,6 +457,21 @@ def get_replacement_path(original_path, mapping):
         fail_reason: if no new_path, more information about failure
     '''
 
+    def get_destination_from_mapping(path):
```
mmorgantaylor, line 460: I generally try to avoid functions-in-functions when I can, though I'm not entirely sure that's a hard and fast rule. But it caught my eye because `mapping` isn't an input to this function... is there a reason not to have `get_destination_from_mapping` defined outside of `get_replacement_path`, and pass `mapping` to it as an input along with `path`?

PR #12, opened 2020-08-04
[DFE-217] [risk=no] Created Script for Creating and Launching the Monitor Submission Workflow

Marianie-Simeon:
## Addresses
https://broadworkbench.atlassian.net/browse/DFE-217

## Updates
Created a script that will create and launch the Monitor Submission Workflow

## Rebased
08/07/2020

### DIFF
--- a/scripts/call_monitor_submission_demo.py
+++ b/scripts/call_monitor_submission_demo.py
@@ -13,6 +13,7 @@ workflow_name = 'monitor_submission'
 workflow = fapi.get_workspace_config(workspace_project, workspace_name, workflow_repo, workflow_name)
 if workflow.status_code != 200:
     print(workflow.content)
+    raise ferrors.FireCloudServerError(workflow.status_code, workflow.content)
 else:
     workflow_json = workflow.json()
 
@@ -21,11 +22,13 @@ else:
     updated_workflow = fapi.update_workspace_config(workspace_project, workspace_name, workflow_repo, workflow_name, workflow_json)
     if updated_workflow.status_code != 200:
         print(updated_workflow.content)
+        raise ferrors.FireCloudServerError(updated_workflow.status_code, updated_workflow.content)
     else:
         # Launching the Updated Monitor Submission Workflow
         create_submisson_response = fapi.create_submission(workspace_project, workspace_name, workflow_repo, workflow_name, entity=None, etype=None, expression=None, use_callcache=True)
         if create_submisson_response.status_code != 201:
             print(create_submisson_response.content)
+            raise ferrors.FireCloudServerError(create_submisson_response.status_code, create_submisson_response.content)
         else:
             print("Successfully Created Submisson")
             print(create_submisson_response.json())
--- a/scripts/call_monitor_submission_demo.py
+++ b/scripts/call_monitor_submission_demo.py
@@ -13,7 +13,6 @@ workflow_name = 'monitor_submission'
 workflow = fapi.get_workspace_config(workspace_project, workspace_name, workflow_repo, workflow_name)
 if workflow.status_code != 200:
     print(workflow.content)
-    raise ferrors.FireCloudServerError(workflow.status_code, workflow.content)
 else:
     workflow_json = workflow.json()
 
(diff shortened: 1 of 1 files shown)


scripts/call_monitor_submission_demo.py
```diff
@@ -0,0 +1,34 @@
...
+from firecloud import api as fapi
+from firecloud import errors as ferrors
+
+workspace_name = 'HCA_Optimus_Pipeline_MS'
+workspace_project = 'broad-firecloud-dsde'
+submission_id = '990e5930-a9b8-4f70-b220-11b7ffe702bb'
+# workflow_name has to be unique everytime
+workflow_name = 'monitor_submission_22'
+
+json_body = {
```
mmorgantaylor, line 11: where did this json come from? can it be retrieved using an API rather than hard-coded in the script? e.g. pretty sure we don't want `"namespace": 'horsefish'` hard coded here

scripts/call_monitor_submission_demo.py
```diff
@@ -0,0 +1,34 @@
...
+    "name": workflow_name,
+    'methodConfigVersion': 14,
+    'methodRepoMethod': {'methodName': 'monitor_submission', 'methodVersion': 14, 'methodNamespace': 'horsefish', 'methodUri': 'agora://horsefish/monitor_submission/14', 'sourceRepo': 'agora'},
+    "inputs": {"monitor_submission.submission_id": f'"{submission_id}"', "monitor_submission.terra_project": f'"{workspace_project}"', "monitor_submission.terra_workspace": f'"{workspace_name}"'},
+    "outputs": {},
+    "prerequisites": {},
+    'deleted': False,
+    'source': 'agora://horsefish/monitor_submission/14'
+}
+create_workflow_response = fapi.create_workspace_config(workspace_project, workspace_name, json_body)
```
mmorgantaylor, line 22: what does create_workspace_config do? could you add a comment here to describe the effect of this call?

scripts/call_monitor_submission_demo.py
```diff
@@ -0,0 +1,34 @@
...
+if create_workflow_response.status_code != 201:
+    print(create_workflow_response.content)
+    raise ferrors.FireCloudServerError(create_workflow_response.status_code, create_workflow_response.content)
+else:
+    print("Successfully Created Workflow")
+    create_submisson_response = fapi.create_submission(workspace_project, workspace_name, 'horsefish', workflow_name, entity=None, etype=None, expression=None, use_callcache=True)
```
mmorgantaylor, line 28: similarly, `'horsefish'` here should not be hard-coded - better to pull it from the json e.g. `json_body['namespace']`

scripts/call_monitor_submission_demo.py
Marianie-Simeon: I made the JSON, with the comments from fiss.apy.py script, input.json, and playing around

scripts/call_monitor_submission_demo.py
Marianie-Simeon: It creates a workflow, I'll add comments

scripts/call_monitor_submission_demo.py
mmorgantaylor: hmm, then can you populate its contents more programmatically? e.g. 'horsefish' shouldn't be hard-coded anywhere, nor should methodConfigVersion be hardcoded as 14, etc etc

scripts/call_monitor_submission_demo.py
```diff
@@ -0,0 +1,34 @@
...
+submission_id = '990e5930-a9b8-4f70-b220-11b7ffe702bb'
+workflow_repo = 'horsefish'
+workflow_name = 'monitor_submission'
+
+# Getting the json for the Monitor Submission Workflow
+workflow = fapi.get_workspace_config(workspace_project, workspace_name, workflow_repo, workflow_name)
+if workflow.status_code != 200:
+    print(workflow.content)
+    raise ferrors.FireCloudServerError(workflow.status_code, workflow.content)
+else:
```
mmorgantaylor, line 17: you don't need this else -- since you raise an error if it's not a success code, you can just continue on with code and you know it will only continue if you got a success code. (same thing for line 26)

mmorgantaylor: and line 32

scripts/call_monitor_submission_demo.py
```diff
@@ -0,0 +1,31 @@
...
+workspace_project = 'broad-firecloud-dsde'
+submission_id = '990e5930-a9b8-4f70-b220-11b7ffe702bb'
+workflow_repo = 'horsefish'
+workflow_name = 'monitor_submission'
+
+# Getting the json for the Monitor Submission Workflow
+workflow = fapi.get_workspace_config(workspace_project, workspace_name, workflow_repo, workflow_name)
+if workflow.status_code != 200:
+    print(workflow.content)
+else:
```
mmorgantaylor, line 16: ah, no, I'm sorry. you do want to keep the raise error, but remove the else statement.

so it should be like:

`if status != errorcode:
    print(whatever)
    raise

[continue with code]
`

Marianie-Simeon: Oh, I understand now

PR #14, opened 2020-08-24
[DFE-237][risk=no] Update monitor_submission

Marianie-Simeon:
Problem statement:

The NeMO team asked for workspace information in the output JSON from our workflow monitoring pipeline.

Solution:

Added workspace information this information to the output JSON created in the Python code.

-Update monitor_submission.py 
    ~ Fixed type error

### DIFF
--- a/scripts/monitor_submission.py
+++ b/scripts/monitor_submission.py
@@ -49,14 +49,11 @@ def monitor_submission(terra_workspace, terra_project, submission_id, sleep_time
             # terra_details to be appended to json
             terra_details = {"terra_workspace": terra_workspace, "terra_project": terra_project}
 
-            # loading JSON submission_metadata:
-            submission_data = json.loads(submission_metadata)
-
-            # appending the terra_details
-            submission_data.update(terra_details)
+            # udating/appending the terra_details to current json
+            submission_metadata.update(terra_details)
 
             # writes final result to file:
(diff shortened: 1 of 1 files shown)


scripts/monitor_submission.py
```diff
@@ -49,14 +49,11 @@ def monitor_submission(terra_workspace, terra_project, submission_id, sleep_time
             # terra_details to be appended to json
             terra_details = {"terra_workspace": terra_workspace, "terra_project": terra_project}
 
-            # loading JSON submission_metadata:
-            submission_data = json.loads(submission_metadata)
-
-            # appending the terra_details
-            submission_data.update(terra_details)
+            # udating/appending the terra_details to current json
```
mmorgantaylor, line 52: typo "updating"

scripts/monitor_submission.py
```diff
@@ -49,14 +49,11 @@ def monitor_submission(terra_workspace, terra_project, submission_id, sleep_time
...
+            submission_metadata.update(terra_details)
```
mmorgantaylor, line 54: you could move the `with open(save_name, 'w') as f:` line to here -- the updating of submission_metadata with terra_details doesn't need to be inside that save block.

scripts/monitor_submission.py
Marianie-Simeon: Thank you!

PR #15, opened 2020-09-14
[DFE-167][risk=no] Creates test to compare warehouse and warehouse_dev tables

Marianie-Simeon:
### **Addresses** 
https://broadworkbench.atlassian.net/jira/software/projects/DFE/boards/62

### **Commit Summary**

- Created compare_warehouse_tables.py
- Outputs if columns and rows of the compare warehouse and warehouse_dev tables match
- Outputs the count and information of the extra rows in dev and prod

### **Output**
  ```
+---+---------------------------+----------------+-------------+-------------+--------------+
|   |        Table_Name         | Columns_Match? | Rows_Match? | In_Dev_Only | In_Prod_Only |
+---+---------------------------+----------------+-------------+-------------+--------------+
| 0 | sam_permissions_postgres  |      True      | True (99%)  |    1493     |    20019     |
| 1 |         terra_nps         |      True      | True (99%)  |      0      |      1       |
| 2 |     rawls_workspaces      |      True      | True (99%)  |      0      |      19      |
| 3 |           email           |      True      | True (99%)  |      3      |     113      |
| 4 |          thurloe          |      True      | True (99%)  |     51      |     1369     |
| 5 |     rawls_access_logs     |      True      | True (99%)  |      0      |     5046     |
| 6 |     rawls_submissions     |      True      | True (99%)  |      0      |      60      |
| 7 |      leo_access_logs      |      True      | True (99%)  |      0      |    62678     |
| 8 |      rawls_workflows      |      True      | True (99%)  |      0      |      94      |
| 9 | orchestration_access_logs |      True      | True (99%)  |      0      |    32033     |
+---+---------------------------+----------------+-------------+-------------+--------------+
More Information:
terra_nps table didn't 100% match:  yesterday : 
rawls_workspaces table didn't 100% match:  yesterday : One Month : 
rawls_access_logs table didn't 100% match:  yesterday : 
rawls_submissions table didn't 100% match:  yesterday : 
leo_access_logs table didn't 100% match:  yesterday : 
rawls_workflows table didn't 100% match:  yesterday : 
orchestration_access_logs table didn't 100% match:  yesterday : 
```

### DIFF
--- a/scripts/compare_warehouse_tables.py
+++ b/scripts/compare_warehouse_tables.py
@@ -1,98 +1,244 @@
-"""WILL FILL OUT."""
-# from googlecloud_bigquery_class import BigQuery
+"""Compare the dev_warehouse with the prod warehouse to make sure that changes to code did not break functionality."""
+from datetime import datetime, timedelta
 from google.cloud import bigquery
-# import pandas as pd
+from tabulate import tabulate
+import pandas as pd
+import argparse
+import warnings
+
+# TODO Setup For Running The Code:
+# 1. Make sure to update warehouse_dev:
+#   A. Go to https://fcprod-jenkins.dsp-techops.broadinstitute.org
+#   B. Go to the Data Warehouse tab (if applicable)
+#   C. Click the WIP-data-warehouse-python-development job
+#   D. Check if code is pulling from */master branch (Info found in Configure Section, Under Source Code Management -> Branches to build)
+#   E. In the WIP-data-warehouse-python-development job, press Build Now
+# 2. Run "gcloud auth login" with firecloud.org email
+# 3. Run "gcloud auth application-default login" with firecloud.org email
+# 4. Run "gcloud auth application-default set-quota-project broad-dsde-dev"
 
 
-def call_bigquery(env_query, project_name):
+def call_bigquery(bq, env_query, project_name, verbose=False):
     """Call BigQuery table."""
-    # Construct a BigQuery client object.
-    bq = bigquery.Client(project=project_name)
-
     # Start job
     query = bq.query(env_query)
-    print("Starting job.")
+    if verbose:
+        print("Starting job.")
 
     # Gather results
     results = query.result()
-    print("Job finished.")
+    if verbose:
+        print("Job finished.")
 
     # read RowIterator object into pandas df
     results_df = results.to_dataframe()
-    print(results_df)
+    if verbose:
+        print(results_df)
     return results_df
 
 
-def get_tables(schema):
+def get_tables(schema, verbose=False):
     """Get a list of the tables in warehouse_dev."""
-    dev_tables = list(schema['table_name'])  # type list
+    # Getting the list of tables
+    dev_tables = list(schema['table_name'])
+
+    # Remove tables not in Prod or too long to parse
     dev_tables.remove("rawls_entity")
     dev_tables.remove("rawls_entity_attribute")
     dev_tables.remove("cromwell_metadata")
-    print(f"List of dev tables in {dev_tables}")
+    if verbose:
+        print(f"List of dev tables in {dev_tables}")
     return dev_tables
 
 
-def get_newrows(table_names):
-    """Get the list of new row in dev."""
-    project = "broad-dsde-prod-analytics-dev"
+def get_newrows(bq, project, table_names, tables_with_timestamps, verbose):
+    """Getting the table and information of new rows in dev and prod."""
+    # Creating the master list of the new rows in dev and prod for the table
+    results_newrows = []
+
+    # Getting today date
+    today = datetime.today()
+
+    # initializing the time count variable that Loops through the dates: today, yesterday, 1 month and six months
+    time_count = 0
+
+    # Initializing the Boolean has_timestamp that checks if the table has a timestamp
+    has_timestamp = False
+
+    # Initializing the more_info variable that will output extra information of the tables
+    more_info = ""
+
+    # Looping through all the table names for columns and rows match, and extra rows in Dev and prod
     for table_name in table_names:
-        print(table_name)
-        query_get_dev_schema = f"""WITH
-        DevData AS (
+        row_result = []
+        time_iteration = True
+        not_matched = False
+        not_matched_info = ""
+        newrows = [table_name]
+        while time_iteration:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: scripts/compare_warehouse_tables.py [hunk 1]
```diff
@@ -0,0 +1,244 @@
...
+
+def get_tables(schema, verbose=False):
+    """Get a list of the tables in warehouse_dev."""
+    # Getting the list of tables
+    dev_tables = list(schema['table_name'])
+
+    # Remove tables not in Prod or too long to parse
+    dev_tables.remove("rawls_entity")
+    dev_tables.remove("rawls_entity_attribute")
+    dev_tables.remove("cromwell_metadata")
```
mmorgantaylor, line 48: this is indeed a huge table, so we don't want to parse it all, but we do need to check it somehow!

scripts/compare_warehouse_tables.py
```diff
@@ -0,0 +1,244 @@
...
+    return results_df
+
+
+def get_tables(schema, verbose=False):
+    """Get a list of the tables in warehouse_dev."""
+    # Getting the list of tables
+    dev_tables = list(schema['table_name'])
+
+    # Remove tables not in Prod or too long to parse
+    dev_tables.remove("rawls_entity")
```
mmorgantaylor, line 46: `rawls_entity` and `rawls_entity_attribute` will likely (hopefully) be included at some point - it might be nice to make a variable like `TABLES_TO_SKIP = ['rawls_entity','rawls_entity_attribute']` and then in this function do a for loop and remove those tables, but also add a comment to the output of the script that notes that these tables were skipped

scripts/compare_warehouse_tables.py
```diff
@@ -0,0 +1,244 @@
...
+    # Looping through all the table names for columns and rows match, and extra rows in Dev and prod
+    for table_name in table_names:
+        row_result = []
+        time_iteration = True
+        not_matched = False
+        not_matched_info = ""
+        newrows = [table_name]
+        while time_iteration:
+            # Checking if table has a timestamps and (if applicable) looping through today, yesterday, one month ago, and six month ago
+            if table_name in tables_with_timestamps:
```
mmorgantaylor, line 80: I don't think we should check "today" at all, since we know it is going to be different - I'd remove this entire block

scripts/compare_warehouse_tables.py
```diff
@@ -0,0 +1,244 @@
...
+                if time_count == 0:
```
mmorgantaylor, line 81: I'm not sure I understand the `if time_count` logic... I think this would be more readable if you fed in a list of timestamps to use and looped through it. e.g. `DAYS_TO_CHECK = [one_day, one_month, six_months]` and then `for day_to_check in DAYS_TO_TEST: ` etc.

scripts/compare_warehouse_tables.py
```diff
@@ -0,0 +1,244 @@
...
+            else:
+                timestamp_DevData = ""
+                timestamp_ProdData = ""
+                time_iteration = False
+                has_timestamp = False
+
+            # Initializing query
+            extra_rows_query = f"""WITH
+            DevData AS (
+                SELECT
```
mmorgantaylor, line 116: so how much data does this query scan (check in BigQuery) for the cromwell_metadata table?

scripts/compare_warehouse_tables.py
```diff
@@ -0,0 +1,244 @@
...
+            newrows.append("True")
+        else:
+            newrows.append("False")
+            more_info += f"{table_name} table columns Not Matched on " + (", ".join(list(list(set(prod_table_columns) - set(dev_table_columns)) + list(set(dev_table_columns) - set(prod_table_columns))))) + " \n"
+
+        # Adding Rows_Match boolean and percentage (if not 100%) to the newrow row
+        if dev_table_rows is None:
+            newrows.append("True")
+        else:
+            match_percent = int(100 - (((dev_table_rows + prod_table_rows) / dev_table.num_rows) * 100))
```
mmorgantaylor, line 176: is this being calculated for each time point? your output only shows one value here, which timepoint is that for?

scripts/compare_warehouse_tables.py
```diff
@@ -0,0 +1,244 @@
...
+            if match_percent > 85:
```
mmorgantaylor, line 177: I don't think we want this cutoff, it either matches or doesn't

scripts/compare_warehouse_tables.py
```diff
@@ -0,0 +1,244 @@
...
+
+    # Getting all the warehouse tables
+    query_get_dev_schema = "SELECT * FROM broad-dsde-prod-analytics-dev.warehouse_dev.INFORMATION_SCHEMA.TABLES"
+    dev_dataset_schema = call_bigquery(bq, query_get_dev_schema, project, verbose)
+
+    # Getting all the warehouse tables names
+    dev_table_list = get_tables(dev_dataset_schema, verbose)
+
+    # Dict with the warehouse tables that has a date value 
+    tables_with_timestamps = {"terra_nps": "timestamp",
```
mmorgantaylor, line 231: it would be nice to pull this programmatically from the schema

mmorgantaylor: (those are in workbench-analytics, which is where you'll want to put this script, not in horsefish!)

scripts/compare_warehouse_tables.py
Marianie-Simeon: 412.7 MB with timestamps

mmorgantaylor: cool, that seems manageable, can you try not removing cromwell_metadata from the list of tables and see if you can run the check on it?

scripts/compare_warehouse_tables.py
```diff
@@ -37,15 +37,15 @@ def call_bigquery(bq, env_query, project_name, verbose=False):
...
     # Getting the list of tables
     dev_tables = list(schema['table_name'])
 
-    # Remove tables not in Prod or too long to parse
-    dev_tables.remove("rawls_entity")
-    dev_tables.remove("rawls_entity_attribute")
-    dev_tables.remove("cromwell_metadata")
+    # Remove tables not in Prod
+    for table in tables_to_skip:
+        dev_tables.remove(table)
```
mmorgantaylor, line 47: very nice!

PR #19, opened 2020-11-05
[DFE-2781] add + remove terra service banner

schaluva:
In the event of a Service Incident related to the Terra UI, SDLC notes that a message must be posted to the platform for users to be made aware of the ongoing issue. The messaging is provided to users of the platform with a modifiable banner.The process of posting the banner has a few steps that are manual but this python script provides a streamlined solution for internal members that are on a suitability roster.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -2,21 +2,34 @@
 
 ### A.  terra_service_banner.py
 #### Description
-    In the event of a Service Incident related to the Terra UI, SDLC notes that a message must be posted to the platform for users to be made aware of the ongoing issue. The messaging is provided to users of the platform with a modifiable banner.The process of posting the banner has a few steps that are manual but this python script provides a streamlined solution for internal members that are on a suitability roster.
+    In the event of a Service Incident affecting the Terra platform, SDLC states that a message must be posted to the platform for users to be made aware of the ongoing issue. The messaging is provided via the UI with a modifiable banner.The process of posting the banner has a few steps that are manual but this python script provides a streamlined solution for internal members that are on a suitability roster.
 
-    The script can be run to post and delete template banners but is also equipped to handle a custom message via a .json file.
+    The script can be run to post and clear the standard service incident banner but is also equipped to handle a custom message via a .json file.
+
+    The code in this script modifies only a single object, mandatorily named "alerts.json", in a specific bucket. The Terra UI monitors the status of the .json file; if there are contents, a banner is posted, but if the .json is empty, the banner is cleared. 
 
 #### Execution
     1. Authenticate user with the email address that is on the suitability roster:
         `gcloud auth login user@email.com`
     2. Execute script:
-        `python3 terra_service_banner.py --env ["prod" or "dev"] [--delete] [--json path to custom banner .json]`
+        `python3 terra_service_banner.py --env ["prod" or "dev"] [--delete] [--json path to custom_banner.json]`
 
-    Eexample:
+    Example:
     1. To post template banner to the production environment:
-        `python3 modify_dph_headers.py --env prod
-    2. To delete any banner (template or custom) from the production environment:
-        `python3 modify_dph_headers.py --env prod --delete`
+        `python3 terra_service_banner.py --env prod`
+    2. To post custom banner to the production environment:
+        `python3 terra_service_banner.py --env prod --json custom_banner.json`
+    3. To clear the banner (template or custom) from the production environment:
+        `python3 terra_service_banner.py --env prod --delete`
+
+    Example custom_banner.json:
+        ```[
+            {
+                "title":"Example Custom Banner: Service Incident",
+                "message":"We are currently investigating an issue impacting the platform. Information about this incident will be made available here.",
+                "link":"https://support.terra.bio/hc/en-us/sections/360003692231-Service-Notifications"
+            }
+        ]```
 
 #### Output
-    If posting the indicated Terra environment will have either a template or custom banner message and if deleting, any banner that is present will be removed.
+    If posting, the indicated Terra environment will have either a template or custom banner message and if deleting, the banner that is present will be removed.
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,22 @@
+## Horsefish Scripts 
+
+### A.  terra_service_banner.py
+#### Description
+    In the event of a Service Incident related to the Terra UI, SDLC notes that a message must be posted to the platform for users to be made aware of the ongoing issue. The messaging is provided to users of the platform with a modifiable banner.The process of posting the banner has a few steps that are manual but this python script provides a streamlined solution for internal members that are on a suitability roster.
+
+    The script can be run to post and delete template banners but is also equipped to handle a custom message via a .json file.
+
+#### Execution
+    1. Authenticate user with the email address that is on the suitability roster:
+        `gcloud auth login user@email.com`
+    2. Execute script:
+        `python3 terra_service_banner.py --env ["prod" or "dev"] [--delete] [--json path to custom banner .json]`
+
+    Eexample:
+    1. To post template banner to the production environment:
+        `python3 modify_dph_headers.py --env prod
+    2. To delete any banner (template or custom) from the production environment:
+        `python3 modify_dph_headers.py --env prod --delete`
+
+#### Output
+    If posting the indicated Terra environment will have either a template or custom banner message and if deleting, any banner that is present will be removed.
--- a/scripts/terra_service_banner.py
+++ b/scripts/terra_service_banner.py
@@ -2,19 +2,37 @@ import argparse
 from google.cloud import storage as gcs
 
 
-def convert_json_to_string(json, env):
+def convert_json_to_string(env, json):
     """Convert json file to string if supplied with custom banner json file."""
 
     # open json file and read contents to a string
     with open(json, "r") as j:
         custom_banner_string = j.read()
 
-    print(f'Starting custom banner upload to {env} Terra.')
     return(custom_banner_string)
 
 
-def update_service_banner(json, env):
+def update_service_banner(env, json_string=None):
     """Push json to bucket in selected production environment."""
+
+    # if no custom banner json is passed in, post standard banner text/string
+    if not json_string:
+        banner_text = """[
+                        {
+                        "title":"Service Incident",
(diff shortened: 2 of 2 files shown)


scripts/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+#### Execution
+    1. Authenticate user with the email address that is on the suitability roster:
+        `gcloud auth login user@email.com`
+    2. Execute script:
+        `python3 terra_service_banner.py --env ["prod" or "dev"] [--delete] [--json path to custom banner .json]`
+
+    Eexample:
+    1. To post template banner to the production environment:
+        `python3 modify_dph_headers.py --env prod
```
mmorgantaylor, line 17: I think you forgot to update the script name here and 2 lines down :)

scripts/terra_service_banner.py
```diff
@@ -0,0 +1,98 @@
+import argparse
+from google.cloud import storage as gcs
+
+
+def convert_json_to_string(json, env):
```
mmorgantaylor, line 5: the name of this function is confusing, since it also calls update_service_banner - I'd recommend separating those functions. if you keep this function named as-is, it would make sense for it to return `custom_banner_string` and then you can call `update_service_banner` in line 91

scripts/terra_service_banner.py
```diff
@@ -0,0 +1,98 @@
...
+    blob.acl.save()
+
+    # set and save OWNER access for suitable_group on json object (gsutil ach ch -g suitable_group:O)
+    blob.acl.group(suitable_group).grant_owner()
+    blob.acl.save()
+
+    print("Banner action complete.")
+
+
+def post_banner(env):
```
mmorgantaylor, line 50: I'd rename this something like `post_standard_banner` to make it clear that this is a basic template banner

scripts/terra_service_banner.py
```diff
@@ -0,0 +1,98 @@
...
+    # push json string to bucket - delete banner
+    print(f'Starting banner removal from {env} Terra.')
+    update_service_banner(banner_text, env)
+
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser(description='Publish or remove a production incident banner on Terra UI.')
+    parser.add_argument('--env', required=True, help='"prod" or "dev" Terra environment for banner.')
+    parser.add_argument('--json', help='set to post custom banner (via json file) to Terra UI.')
```
mmorgantaylor, line 82: what's the type of this? string? I'd add that (`type=string`) to this and to the env arg

scripts/terra_service_banner.py
```diff
@@ -0,0 +1,98 @@
...
+                        "link":"https://support.terra.bio/hc/en-us/sections/360003692231-Service-Notifications"
+                        }
+                    ]"""
+
+    # push json string to bucket - post banner
+    print(f'Starting template banner upload to {env} Terra.')
+    update_service_banner(banner_text, env)
+
+
+def delete_banner(env):
```
mmorgantaylor, line 67: - is it the case that only one banner can ever be up on Terra? 
if not: 
- does this function remove all banners? would we ever need to remove one but keep another?

scripts/README.md
```diff
@@ -0,0 +1,22 @@
...
+#### Description
+    In the event of a Service Incident related to the Terra UI, SDLC notes that a message must be posted to the platform for users to be made aware of the ongoing issue. The messaging is provided to users of the platform with a modifiable banner.The process of posting the banner has a few steps that are manual but this python script provides a streamlined solution for internal members that are on a suitability roster.
+
+    The script can be run to post and delete template banners but is also equipped to handle a custom message via a .json file.
+
+#### Execution
+    1. Authenticate user with the email address that is on the suitability roster:
+        `gcloud auth login user@email.com`
+    2. Execute script:
+        `python3 terra_service_banner.py --env ["prod" or "dev"] [--delete] [--json path to custom banner .json]`
```
mmorgantaylor, line 13: could you add an example of the contents of the json file that this would need?

scripts/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+    Eexample:
+    1. To post template banner to the production environment:
+        `python3 modify_dph_headers.py --env prod
+    2. To delete any banner (template or custom) from the production environment:
```
mmorgantaylor, line 18: see comment/questions below, but what does "any" mean? "all banners", right?

scripts/README.md
```diff
@@ -0,0 +1,22 @@
...
+        `python3 modify_dph_headers.py --env prod --delete`
+
+#### Output
+    If posting the indicated Terra environment will have either a template or custom banner message and if deleting, any banner that is present will be removed.
```
mmorgantaylor, line 22: ah I think it would be clearer to say "all banners will be removed" rather than "any" - maybe a small detail but for me it helps with clarity. 

do you want to add the possibility of removing a specific banner?

scripts/terra_service_banner.py
schaluva: For the sole purpose of service incidents, this script acts on only a single object (specifically named "alerts.json") in a specific bucket. It only modifies the contents of that .json and Terra UI reads from that object. If there are contents, it posts a banner, and if its empty, it shows no banner (aka delete banner). There are no other banners on Terra that Frontline or anyone using this script could activate/modify etc. The other things that might pop up are errors that appear in the same shape/form as a banner but are not banners. The only thing someone can take down (delete) is the same banner that is put up via that "alerts.json" object.

mmorgantaylor: got it, that makes sense, thanks! maybe worth including that explanation in the readme, just for someone new to the whole thing?

and to me that makes this function more like `clear_all_service_banners` but that's a personal preference!

scripts/README.md
schaluva: "any" was intended to be either the custom or template banner but really those words describe the content of the banner, it doesnt really mean the banner is different so I can change the language!

scripts/terra_service_banner.py
schaluva: totally! thats great feedback to have for someone that is new to the process! ill make the function different

scripts/terra_service_banner.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def convert_json_to_string(json, env):
+    """Convert json file to string if supplied with custom banner json file."""
+
+    # open json file and read contents to a string
+    with open(json, "r") as j:
+        custom_banner_string = j.read()
+
+    print(f'Starting custom banner upload to {env} Terra.')
```
mmorgantaylor, line 12: this print statement probably belongs somewhere else?

scripts/terra_service_banner.py
```diff
@@ -0,0 +1,99 @@
...
+    return(custom_banner_string)
+
+
+def update_service_banner(json, env):
```
mmorgantaylor, line 16: you could make the `json` input optional (so `update_service_banner(env, json=None)` and then if json is None, make it the default text - that would eliminate the need for `post_standard_banner` entirely

scripts/README.md
```diff
@@ -0,0 +1,35 @@
+## Horsefish Scripts 
+
+### A.  terra_service_banner.py
+#### Description
+    In the event of a Service Incident affecting the Terra platform, SDLC states that a message must be posted to the platform for users to be made aware of the ongoing issue. The messaging is provided via the UI with a modifiable banner.The process of posting the banner has a few steps that are manual but this python script provides a streamlined solution for internal members that are on a suitability roster.
+
+    The script can be run to post and clear the standard service incident banner but is also equipped to handle a custom message via a .json file.
+
+    The code in this script modifies only a single object, mandatorily named "alerts.json", in a specific bucket. The Terra UI monitors the status of the .json file; if there are contents, a banner is posted, but if the .json is empty, the banner is cleared.
```
mmorgantaylor, line 9: this is super clear, thank you!!

scripts/terra_service_banner.py
schaluva: cool, i combined the two and added print statements into the if/else block depending on the json_string that is passed in, if any.

scripts/terra_service_banner.py
```diff
@@ -0,0 +1,100 @@
...
+
+    # open json file and read contents to a string
+    with open(json, "r") as j:
+        custom_banner_string = j.read()
+
+    return(custom_banner_string)
+
+
+def update_service_banner(env, json_string=None):
+    """Push json to bucket in selected production environment."""
```
mmorgantaylor, line 16: v minor: remove the word "production" ?

scripts/terra_service_banner.py
```diff
@@ -0,0 +1,100 @@
...
+
+    # push json string to bucket - clear banner
+    update_service_banner(env, clear_banner_text)
+
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser(description='Publish or remove a production incident banner on Terra UI.')
+    parser.add_argument('--env', type=str, required=True, help='"prod" or "dev" Terra environment for banner.')
+    parser.add_argument('--json', type=str, required=False, help='set to post custom banner (via json file) to Terra UI.')
```
mmorgantaylor, line 82: minor: description is a little confusing - you need to not only set this flag but also give the actual file path to the custom banner json file, right? so something like "path to json containing custom banner information" seems more descriptive to me

scripts/terra_service_banner.py
```diff
@@ -0,0 +1,100 @@
...
+    blob.acl.save()
+
+    print("Banner action complete.")
+
+
+def clear_service_banner(env):
+    """Create json string for upload to GCS location to clear/remove banner."""
+
+    # template json text for banner deletion
+    clear_banner_text = """[]"""
```
mmorgantaylor, line 72: just FYI you don't need the triple quotes here. it's not a problem to have them, but they're typically used for multi-line strings

mmorgantaylor (approved):
this looks great! a few very minor suggestions that you can take or leave.

scripts/terra_service_banner.py
schaluva: ugh yeah my OCD got the best of me and I wanted it to "match" the multi line string for the template banner.

scripts/terra_service_banner.py
schaluva: yep what i had was not clear at all! changed language

scripts/terra_service_banner.py
mmorgantaylor: ha totally valid!

PR #21, opened 2020-12-03
modify banner script to run in jenkins

jacmrob:
- clean up inputs to take strings for banner fields rather than json banner text; build json in-script
- update some packages and aspects of the docker build so it works

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,8 @@ FROM google/cloud-sdk:295.0.0
 ENV HOME=/
 
 # install python3 to make that also available - modified from https://tecadmin.net/install-python-3-7-on-ubuntu-linuxmint/
-RUN apt-get install -y build-essential checkinstall \
+RUN apt-get update \
+    && apt-get install -y build-essential checkinstall \
     && apt-get install -y libreadline-gplv2-dev libncursesw5-dev libssl-dev libsqlite3-dev \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -4,7 +4,8 @@ FROM google/cloud-sdk:295.0.0
 ENV HOME=/
 
 # install python3 to make that also available - modified from https://tecadmin.net/install-python-3-7-on-ubuntu-linuxmint/
-RUN apt-get install -y build-essential checkinstall \
+RUN apt-get update \
```
jacmrob, line 7: I found that when building a fresh image, the build was broken here, likely because we're using an old `apt-get`.  this probably hasn't been caught because builds of `latest` are using caching so it rarely has to rebuild this layer.  adding `apt-get update` seemed to fix things

requirements.txt
```diff
@@ -1,4 +1,5 @@
 firecloud==0.16.25
 pip==19.3.1
 tenacity==5.1.1
-pandas==0.25.3
+pandas==1.1.4
```
jacmrob, line 4: I'm not sure if an older version of pandas was being used deliberately, but I found it was also broken when running a fresh build (again, probably missed because of layer caching).  updating the version fixed things.

schaluva (approved):

mmorgantaylor (approved):
thanks Jackie!

mmorgantaylor (approved):
looks good to me!

PR #22, opened 2020-12-08
Expand the setup instructions for the Cloud Functions example.

deflaux:
Also:
* added an example WDL and entity set to provide more context
* allow users to specify which entity set to use
* facilitate faster testing by authenticating locally as the service account
* refactor more of what was in main into utils
* rename parameters for clarity
* add pydocs

### DIFF
--- a/scripts/launch_workflow_cf/README.md
+++ b/scripts/launch_workflow_cf/README.md
@@ -1,28 +1,122 @@
-## launch_workflow CF
-The scripts in this directory can be used to deploy a Google Cloud Function (CF) that launches a Terra workflow, triggered off of a file being uploaded to a GCP bucket.
+# launch_workflow Cloud Function
 
-Before running `deploy_local.sh`, you must perform these manual steps:
-1. Create or identify the GCP project where this CF will live / be billed  
-  a. Enable the Cloud Functions API by visiting `https://console.developers.google.com/apis/library/cloudfunctions.googleapis.com?project=<your_project_name>`  
-  b. Enable the Cloud Build API by visiting `https://console.developers.google.com/apis/library/cloudbuild.googleapis.com?project=<your_project_name>`
-  c. Enable IAM Service Account Credentials API by visiting `https://console.developers.google.com/apis/library/iamcredentials.googleapis.com?project=<your_project_name>`
-  d. Enable Secret Manager `https://console.cloud.google.com/marketplace/product/google/secretmanager.googleapis.com`
-2. Create or identify the bucket in that project that will be used for files to trigger the CF, e.g. `launch-workflow-test-input`
-3. Create a GCP service account 
-  a. create a SA key (json file), upload it to Secret Manager with the name `service_account_key`, and grant the SA `Secret Manager Secret Accessor` permissions
-4. Register the service account in Terra -  see https://github.com/broadinstitute/terra-tools/tree/master/scripts/register_service_account
-5. Retrieve the proxy group for the service account using this swagger endpoint: https://api.firecloud.org/#/Profile/getProxyGroup
-    a. Grant that SA's proxy group `Storage Object Viewer` permissions to the input bucket
-    b. If your workflow uses a private container image, also grant the SA's proxy group `Storage Object Viewer` permissions to the `artifacts.<project-id>.appspot.com` bucket
-6. Share the Terra workspace with the SA, granting Writer & compute access.
-7. Update the `config.yaml` file in this directory with the Terra workspace and workflow information you want to use, including the name of the workflow input that needs to be updated with the file information from the file that triggers each CF run.
+The scripts in this directory can be used to deploy a Google Cloud Function (CF) that launches a Terra workflow. The Cloud Function is triggered when a file is created or modified in a specific Google Cloud Storage bucket.
 
-These values will be used as input parameters to the deployment script.
+Note: this example is specificaly written assumming:
 
-Usage:
+* the triggering file is an input parameter to the Terra workflow
+* the most recently created entity set should be used, or override this default via `ENTITY_SET_TO_USE`
+* the Terra UI is used to control *all other aspects* of the workflow configuration, such as:
+  * which version of the method is to be used
+  * whether call caching is enabled
+  * whether to delete intermediate outputs
+  * the type of the root entity
+  * how the columns of the root entity map onto the workflow parameters
 
-```./deploy_local.sh PROJECT BUCKET SERVICE_ACCOUNT```
+**New to Cloud Functions and Secret Manager?** Don't start here. Instead see:
 
+* [Python Quickstart | Cloud Functions Documentation | Google Cloud](http://cloud/functions/docs/quickstart-python)
+* [Quickstart | Secret Manager Documentation | Google Cloud](http://cloud/secret-manager/docs/quickstart)
+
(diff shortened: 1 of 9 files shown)


deflaux:
Thank you @mmorgantaylor - I learned a lot from this example!!!

scripts/launch_workflow_cf/README.md
```diff
@@ -1,28 +1,122 @@
...
+1. Create a new [service account](https://console.cloud.google.com/iam-admin/serviceaccounts) within that project. It will look like `<your service account name>@<your project id>.iam.gserviceaccount.com`.
+1. Download a JSON key for the newly created service account and store it in a safe place.
+1. Upload the JSON key to [Secret Manager](https://console.cloud.google.com/security/secret-manager).
+1. For the newly created secret, in column "Actions", choose "Copy Resource ID". You'll need this value for the deployment step and it will look like: `projects/<your project number>/secrets/<your secret name>/versions/1`.
+1. Give the service account the needed permissions to items within the project.
+    1. Grant role `Secret Manager Secret Accessor` to the service account.
+    1. Grant role `Storage Object Viewer` to the service account on the bucket that will trigger the Cloud Function.
+1. [Register the service account](https://github.com/broadinstitute/terra-tools/tree/master/scripts/register_service_account) in Terra.
+1. [Create a new Terra group](https://app.terra.bio/#groups) to hold only the service account. It will look like `
```
mmorgantaylor, line 52: genius to create a Terra group for this rather than having to find the proxy group. this readme is *incredible*, Nicole, wow.

scripts/launch_workflow_cf/README.md
```diff
@@ -1,28 +1,122 @@
...
+<your group name>@firecloud.org`.
+1. Give the Terra Group the needed permissions to items within the project.
+    1. Add the Terra group to the billing project (a.k.a. workspace namespace).
+    1. Grant role `Storage Object Viewer` to the Terra group on the bucket that will trigger the Cloud Function. This is because, presumably, the workflow will need to read the contents of the file that triggered the workflow.
+    1. If your workflow uses a private Google Container Repository image, also grant role `Storage Object Viewer` to the Terra group on the bucket that holds the image (e.g., `artifacts.<your project id>.appspot.com`).
+1. Give the Terra Group the needed permissions within Terra.
+    1. Share the Terra workspace with the Terra group, granting `Writer` and `Can compute` access.
+    1. If the workspace has an Authorization Domain, ensure that the Terra group is a member of the Authorization Domain.
+1. If the workflow method is private, share the workflow method. Note that for the Broad Methods repository, sharing with groups is not yet supported. Instead share the method with the service account.
```
mmorgantaylor, line 61: in this case would you need to share it with the proxy as well? I'm really not sure, I'm still learning when the pet is used for permissions vs when the direct user is used.

scripts/launch_workflow_cf/main.py
```diff
@@ -1,103 +1,68 @@
...
-def launch_workflow(data, context):
-    # extract file information from the triggering PubSub message
-    file_name = data.get('name')
-    bucket_name = data.get('bucket')
-    file_path = f"gs://{bucket_name}/{file_name}"
-
-    print(f"input file: {file_name}; full path: {file_path}")
-
-    prepare_and_launch(file_path)
+import tensorflow as tf
```
mmorgantaylor, line 4: why is tensorflow imported?

scripts/launch_workflow_cf/utils.py
```diff
@@ -5,94 +5,168 @@
...
-
-### Firecloud API calls
+  if response.status_code != success_code:
+    print(response.content)
+    raise FireCloudServerError(response.status_code, response.content)
 
-def get_workspace_config(namespace, workspace, cnamespace, config, headers):
-    uri = f"https://api.firecloud.org/api/workspaces/{namespace}/{workspace}/method_configs/{cnamespace}/{config}"
-    return requests.get(uri, headers=headers)
+def get_workflow_method_config(workspace_namespace, workspace_name, method_namespace, method_name, headers):
```
mmorgantaylor, line 146: my goodness, thank you for catching and fixing the completely inaccurate function names!

scripts/launch_workflow_cf/utils.py
```diff
@@ -5,94 +5,168 @@
...
 
-def get_entities(namespace, workspace, etype, headers):
-    uri = f"https://api.firecloud.org/api/workspaces/{namespace}/{workspace}/entities/{etype}"
-    return requests.get(uri, headers=headers)
+def get_entities(workspace_namespace, workspace_name, etype, headers):
+  uri = f"https://api.firecloud.org/api/workspaces/{workspace_namespace}/{workspace_name}/entities/{etype}"
+  return requests.get(uri, headers=headers)
 
-def create_submission(wnamespace, workspace, cnamespace, config, headers,
+def create_submission(workspace_namespace, workspace_name, method_namespace, method_name, headers,
```
mmorgantaylor, line 158: and for clarifying and fixing input variable names!

scripts/launch_workflow_cf/utils.py
```diff
@@ -5,94 +5,168 @@
...
+
+  # Merge the workflow inputs configured in the Terra UI with the items from 'workflow_parameters'
+  # in-place, overwriting existing keys.
+  workflow_config_json = workflow_config_response.json()
+  workflow_config_json["inputs"].update(workflow_parameters)
+  if not "rootEntityType" in workflow_config_json:
+    raise ValueError("""The current default workflow configuration does not have a root entity type specified.
+      Be sure to use the Terra UI at least once to successfully run the workflow with an entity set.""")
+
+  # Configure the entity set to be used.
```
mmorgantaylor, line 63: seems like there's a ton of whitespace changes - are your tabs 2 spaces rather than 4?

mmorgantaylor: ah Google prefers 2 spaces, while PEP8 recommends 4. we use 4 spaces in all our codebase - would you be okay keeping it at 4 spaces?

scripts/launch_workflow_cf/main.py
deflaux: good catch! not needed at all

scripts/launch_workflow_cf/README.md
deflaux: It appears that the answer is "no" because it works fine shared only with the service account user.

scripts/launch_workflow_cf/utils.py
deflaux: Yes, I ran a code linter on it with a setting for 2 spaces. No problem, I'll switch it back to 4.

This is what I used:
* https://github.com/broadinstitute/ml4h/blob/master/CONTRIBUTING.md#install-pylint
* https://github.com/broadinstitute/ml4h/blob/master/pylintrc
* with a changed value for `indent-string='    '` so that it was 4 spaces instead of 2

deflaux (commented):
Thanks for the review @mmorgantaylor !

scripts/launch_workflow_cf/utils.py
mmorgantaylor: lovely, thanks so much!

mmorgantaylor (approved):
awesome, thanks so much for this great work, Nicole! this is vastly improved.

PR #25, opened 2021-01-28
update to Requester Pays script: use org-wide role that works for all projects

mmorgantaylor:
the change is to add an organization-wide role, "RequesterPaysToggler", that lives in the firecloud.org organization, that can be used with any project inside that org. the previous role used in this script wasn't available in all Terra projects and resulted in an error.

also decreased the wait time, since it'll involve manual re-running pretty much no matter what.

### DIFF
--- a/scripts/set_requester_pays.sh
+++ b/scripts/set_requester_pays.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 
 if (( $# < 1 )); then
-  echo "Usage: $0 PROJECT_ID [PATH_TO_BUCKET_LIST_FILE]"
+  echo "Usage: $0 TERRA_PROJECT_ID [PATH_TO_TERRA_BUCKET_LIST_FILE]"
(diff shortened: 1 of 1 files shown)


scripts/set_requester_pays.sh
```diff
@@ -15,23 +15,23 @@ fi
 PROJECT_ID=$1
 USER_EMAIL=$(gcloud config get-value account)
 MEMBER="user:${USER_EMAIL}"
-ROLE="projects/${PROJECT_ID}/roles/RequesterPays"
+ROLE="organizations/386193000800/roles/RequesterPaysToggler"
```
Marianie-Simeon, line 18: Do we need to have an organizations_id, instead of having it there?

Marianie-Simeon (approved):

scripts/set_requester_pays.sh
mmorgantaylor: that's a good question... i think as long as this is for Terra, that's the organization id - but I can make that more explicit here. thank you!

PR #29, opened 2021-02-24
[DFE-409] anvil_tools

schaluva:
1. created `anvil_tools` directory to capture all anvil scripts that we have written
2. reorganized the README.md and consolidated multiple files into one for all anvil_tools
3. script `set_up_anvil_workspaces.py` to create workspaces with given auth domains and add ACLs. Full details in [DFE-409 JIRA ticket](https://broadworkbench.atlassian.net/jira/software/projects/DFE/boards/62?selectedIssue=DFE-409).

4. add to requirements.txt to be able to run scripts from within docker.


scripts/anvil_tools/README.md
```diff
@@ -0,0 +1,101 @@
+# AnVIL Tools/Scripts
+
+------------------------
+#### This directory contains scripts/tools to automate steps in the AnVIL data delivery workspace creation process.
+------------------------
+
+### Setup
+    Follow the steps below to set up permissions prior to running any of the listed scripts:
+    1. `gcloud auth login username@broadinstitute.org`
+    2. `gcloud auth application-default login`
```
mmorgantaylor, line 10: not necessary, but if desired, you can streamline this by combining to one line: `gcloud auth login username@broadinstitute.org; gcloud auth application-default login`

scripts/anvil_tools/README.md
```diff
@@ -0,0 +1,101 @@
...
+#### **add_user_to_workspace.py**
+##### Description
+    Update/add user/group to workspace ACL (as READER, WRITER, OWNER). 
+    
+    Input is a .tsv file with 3 columns:
+        1. "workspace_name"
+        2. "email"
+        3. "accessLevel" - (READER, WRITER, or OWNER)
+    Output:
+        NA
```
mmorgantaylor, line 24: is there a success code?

scripts/anvil_tools/README.md
```diff
@@ -0,0 +1,101 @@
...
+    1. `--tsv`, `-t`: input .tsv file (required)
+    2. `--project`, `-p`: workspace project/namespace for listed workspaces in tsv (default = anvil_datastorage)
+
+
+#### **post_workspace_attributes.py**
+##### Description
+    Post dataset attributes to workspaces.
+    
+    Input is a .tsv file:
+        1. Template input.tsv linked [here]()
```
mmorgantaylor, line 89: TODO: add link?

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,255 @@
...
+    # create df for output tsv file
+    col_names = ["input_workspace_name", "input_auth_domain_name",
+                 "auth_domain_email", "auth_domain_creation_error",
+                 "workspace_link", "workspace_creation_error",
+                 "workspace_ACLs", "workspace_ACLs_error",
+                 "workspace_setup_status"]
+    output_df = pd.DataFrame(columns=col_names)
+
+    # per row in tsv/df
+    for row in setup_info_df.index:
```
mmorgantaylor, line 200: it would probably be nice to refactor most of what's in this for loop into a `setup_workspace` function that does all the stuff for a single workspace - that's the core of the work that's happening, and `setup_workspaces` is really just a wrapper to apply this to all the workspaces/rows in the input file

mmorgantaylor: the function would take in all workspace-related inputs, and it would return all the info needed to write to your status output file

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,255 @@
...
+        # workspace
+        workspace_name = setup_info_df.loc[row].get(key='workspace_name')
+        output_df.at[row, "input_workspace_name"] = workspace_name
+        create_ws_request = make_create_workspace_request(workspace_name, auth_domain_name, project)  # get json for createWorkspace
+
+        create_workspace_status = create_workspace(create_ws_request, workspace_name, project)  # createWorkspace
+
+        # created or user wants to continue with existing
+        if create_workspace_status in [201, "Y"]:
+            output_df.at[row, "workspace_link"] = f"https://app.terra.bio/#workspaces/{project}/{workspace_name}"
```
mmorgantaylor, line 223: note that if `workspace_name` has spaces in it, you need to replace each space with `%20` for the url

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,255 @@
...
+    output_df = pd.DataFrame(columns=col_names)
+
+    # per row in tsv/df
+    for row in setup_info_df.index:
+        # authorization domain
+        auth_domain_name = setup_info_df.loc[row].get(key='auth_domain_name')
+        output_df.at[row, "input_auth_domain_name"] = auth_domain_name
+        auth_status = create_auth_domain(auth_domain_name)
+
+        if auth_status != 201:  # if auth creation fail
```
mmorgantaylor, line 206: can you rename to be auth_domain_status or ad_status (and similarly change comment)? "auth" has a different meaning than "auth domain" and that confused me for a sec

scripts/anvil_tools/set_up_anvil_workspaces.py
mmorgantaylor, line 200: this is a kind of clunky way of iterating through the rows of a dataframe - and having to do `setup_info_df.loc[row].get(key='auth_domain_name'` each time you want to access a new field in the row is really bulky.

instead you could do
`for index, row in setup_info_df.iterrows():` 
and then you can access columns simply like 
`row['auth_domain_name']`

mmorgantaylor: (if you do this, then the places below where you're using 'row' you'd actually switch to 'index' - e.g. `output_df.at[row, "field"] = whatever` would become `output_df.at[index, "field"] = whatever`)

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,255 @@
...
+        else:
+            output_df.at[row, "workspace_creation_error"] = create_workspace_status  # write error response to col
+            output_df.at[row, "workspace_ACLs_error"] = "Incomplete"  # adding members function not called
+            output_df.at[row, "workspace_setup_status"] = "Failed"
+            continue
+
+    timestamp = datetime.datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+    output_df.to_csv(f"{timestamp}_workspaces_setup_status.tsv", sep="\t", index=False)
+
+    print("If applicable, workspace creation and set-up complete. Check output file for full details.")
```
mmorgantaylor, line 242: could you count the number of successes and report that? like `10/10 workspaces created and set up successfully` ? so that if/when it all does go well you don't have to look at the setup tsv?

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,255 @@
...
+    create_ws_request["name"] = workspace_name
+    create_ws_request["authorizationDomain"] = [{"membersGroupName": f'{auth_domain_name}'}]
+    create_ws_request["attributes"] = {}
+    # TODO: set noWorkspaceOwner = True for data delivery workspaces - picard svc is the only owner
+    create_ws_request["noWorkspaceOwner"] = False
+
+    return create_ws_request
+
+
+def add_members_to_auth_domain(auth_domain_name):
```
mmorgantaylor, line 132: if this is only adding admins, can you rename to something like `add_anvil_admins_to_auth_domain` ? alternatively (this may be better but it's up to you) if you want to keep it general, then i'd include an input list of emails to add, and hardcode the anvil admin email one level up

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,255 @@
...
+    else:
+        print(f"Successfully updated {project}/{workspace_name} with the following user(s)/group(s): {emails}.")
+        return status_code, emails
+
+
+def check_workspace_exists(workspace_name, project="anvil-datastorage"):
+    """Determine if a workspace of given namespace/name already exists."""
+
+    # don't need full response - could be very large and time consuming
+    uri = f"https://api.firecloud.org/api/workspaces/{project}/{workspace_name}?fields=owners,workspace.createdBy,workspace.authorizationDomain"
```
mmorgantaylor, line 61: you might need to parse spaces here too, i'd check on that

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,255 @@
...
+    response = requests.get(uri, headers=headers)
+    status_code = response.status_code
+
+    if status_code == 404:  # if workspace does not exist
+        return False
+    elif status_code == 200:  # if workspace already exists
+        print(f"Workspace already exists with name: {project}/{workspace_name}.")
+        print(f"Existing workspace details: {json.dumps(response.json(), indent=2)}")
+        # make user decide if they want to update/overwrite existing workspace
+        update_existing_ws = input("Would you like to continue modifying the existing workspace? (Y/N)" + "\n")
```
mmorgantaylor, line 77: great solution!! love this

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,255 @@
...
+        return update_existing_ws.upper()
```
mmorgantaylor, line 78: if the user says 'Y', won't `if not status_ws_exists` still evaluate to False and the code will not continue past line 90? if you want this to evaluate to a boolean, you could instead here return `update_existing_ws.upper() == 'N'` and that will return False if they do want to continue, or True if they don't. double check that logic but I think you need something like that?

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+def create_workspace(json_request, workspace_name, project="anvil-datastorage"):
+    """Create the Terra workspace with given authorization domain."""
+
+    # check if workspace already exists
+    status_ws_exists = check_workspace_exists(workspace_name, project)
+
+    # if workspace does not exist, create workspace
+    if not status_ws_exists:
```
mmorgantaylor, line 90: this logic broke my brain a little... `status_ws_exists` could be False or a string or a json.... could you simplify `check_workspace_exists` to simply return True or False, and if it returns True, you prompt the user to continue or not inside THIS function? I think that would help with the logic flow

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,255 @@
...
+            member_status_code, member_response = add_members_to_workspace(workspace_name, auth_domain_name, project)
+            if member_status_code == 200:
+                output_df.at[row, "workspace_ACLs"] = member_response  # add emails to df
+                output_df.at[row, "workspace_setup_status"] = "Success"  # denotes final workspace setup step
+            else:
+                output_df.at[row, "workspace_ACLs_error"] = member_response
+                output_df.at[row, "workspace_setup_status"] = "Failed"
+        # if "N" (user does not want to continue with existing workspace) or other error
+        else:
+            output_df.at[row, "workspace_creation_error"] = create_workspace_status  # write error response to col
```
mmorgantaylor, line 234: as it is, if the user said "N" to "do you want to proceed with this existing workspace", you'll just write "N" to this column. you might consider a more explanatory "error" message in this case (probably something you want to assign farther up)

scripts/anvil_tools/README.md
schaluva: removed output blocks when script writes success/fail to stdout. these scripts will eventually be modified to write an output file and will add back in output sections then.

mmorgantaylor: 👍

scripts/anvil_tools/set_up_anvil_workspaces.py
schaluva: If a user selects "N", it returns a specific message saying that the user chose not to overwrite the existing workspace! Its the create_workspace() function now but it was earlier in check_workspace_exists()

mmorgantaylor: :facepalm: I totally missed that! it's a little confusing that "Y" stays "Y" but "N" gets converted to the message... but this is ultimately a good outcome, so cool!

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,296 @@
...
+    auth_domain_status = create_auth_domain(auth_domain_name)
+
+    if auth_domain_status != 201:  # AD creation fail - abort setup
+        workspace_dict["auth_domain_creation_error"] = auth_domain_status  # update error in dict
+    else:                          # AD creation success - continue setup
+        workspace_dict["auth_domain_email"] = f"{auth_domain_name}@firecloud.org"
+        anvil_admin_group_name = "anvil-admins@firecloud.org"
+        add_member_status = add_member_to_auth_domain(auth_domain_name, anvil_admin_group_name, "ADMIN")
+        if add_member_status != 204:  # adding member to AD fail - abort setup
+            workspace_dict["add_email_to_AD_error"] = auth_domain_status  # update error in dict
```
mmorgantaylor, line 235: similarly here - `return workspace_dict` and don't need the else afterwards

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,296 @@
...
+        else:                         # adding member to AD success - continue setup
+            workspace_dict["email_added_to_AD"] = anvil_admin_group_name
+
+            # workspace creation if AD set up succeeds
+            workspace_name = workspace["workspace_name"].replace(" ", "%20")
```
mmorgantaylor, line 240: I'd name this something like workspace_name_no_spaces

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,296 @@
...
+            workspace_dict["input_workspace_name"] = workspace["workspace_name"]
+            create_ws_request = make_create_workspace_request(workspace_name, auth_domain_name, project)  # json for API request
```
mmorgantaylor, line 242: I don't think you want to put the workspace_name_no_spaces into this function - then the name of the workspace is going to have `%20` in it.

scripts/anvil_tools/set_up_anvil_workspaces.py
mmorgantaylor: actually I'd just make a helper function that's something like `de_spacify_string` that you can just call when you need to do this. you don't want workspace_name itself to have `%20` in it

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,296 @@
...
+
+            create_workspace_status = create_workspace(create_ws_request, workspace_name, project)  # create workspace
+
+            if create_workspace_status in [201, 200]:  # new ws created or user selected 'continue with existing ws'
+                workspace_dict["workspace_link"] = f"https://app.terra.bio/#workspaces/{project}/{workspace_name}"
+                # add member ACLs to workspace
+                member_status_code, member_response = add_members_to_workspace(workspace_name, auth_domain_name, project)
+                if member_status_code == 200:  # adding ACLs to workspace success
+                    workspace_dict["workspace_ACLs"] = member_response  # add emails to df
+                    workspace_dict["workspace_setup_status"] = "Success"  # final workspace setup step
```
mmorgantaylor, line 252: here as well you can return workpace_dict, remove else

mmorgantaylor (changes requested):

scripts/anvil_tools/set_up_anvil_workspaces.py
mmorgantaylor: you said the api doesn't need spaces removed, so i'll resolve this

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,305 @@
...
+        workspace_dict["add_email_to_AD_error"] = auth_domain_status  # update error in dict
+        return workspace_dict
+
+    # adding member to AD success - continue setup
+    workspace_dict["email_added_to_AD"] = anvil_admin_group_name  # update dict with member added to AD
+
+    # workspace creation if AD set up succeeds
+    workspace_name = workspace["workspace_name"]
+    workspace_dict["input_workspace_name"] = workspace_name
+    create_ws_request = make_create_workspace_request(workspace_name, auth_domain_name, project)  # json for API request
```
mmorgantaylor, line 248: I'd move this entire call to inside `create_workspace` - you don't need to create the json before calling that function. it would allow you to make the inputs to `create_workspace` more intuitive (it's sort of strange for it to take in a json-formatted-request from the start - that's a hard thing to program around if you're going to reuse it) - you could have it take in workspace_name, project, and auth_domain_name

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,305 @@
...
+        member_status_code, member_response = add_members_to_workspace(workspace_name, auth_domain_name, project)
+
+        if member_status_code != 200:  # adding ACLs to workspace fail
+            workspace_dict["workspace_ACLs_error"] = member_response
+            return workspace_dict
+
+        # adding ACLs to workspace success
+        workspace_dict["workspace_ACLs"] = member_response  # update dict with ACL emails
+        workspace_dict["workspace_setup_status"] = "Success"  # final workspace setup step
+    else:  # if "N" (user does not want to continue with existing workspace) or other error
```
mmorgantaylor, line 264: this is small - but to be consistent with your previous code, where you use the `if` statements to check for errors, you could flip this around and on line 252 say 
```
if create_workspace_status not in [201, 200]: 
    workspace_dict["workspace_creation_error"] = create_workspace_status  # update error in dict
    return workspace_dict
```
and then proceed with the rows 253 onward (again not in an else statement)

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,309 @@
...
+        return workspace_dict
+
+    # AD creation success - continue setup
+    workspace_dict["auth_domain_email"] = f"{auth_domain_name}@firecloud.org"  # update dict with AD email
+    # add members to AD
+    anvil_admin_group_name = "anvil-admins@firecloud.org"
+    add_member_status = add_member_to_auth_domain(auth_domain_name, anvil_admin_group_name, "ADMIN")
+
+    if add_member_status != 204:  # adding member to AD fail - abort setup
+        workspace_dict["add_email_to_AD_error"] = auth_domain_status  # update error in dict
```
mmorgantaylor, line 242: did you mean add_member_status here? (but also i think you should refactor this :)

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,309 @@
...
+    # -H  "accept: application/json" -H  "Authorization: Bearer [token]
+    headers = {"Authorization": "Bearer " + get_access_token(), "accept": "application/json"}
+
+    # capture response from API and parse out status code
+    response = requests.get(uri, headers=headers)
+    status_code = response.status_code
+
+    if status_code == 404:          # if workspace does not exist
+        return False, None, None
+    elif status_code == 200:        # if workspace already exists
```
mmorgantaylor, line 90: here you can just do `if status_code == 200:` and 2 lines down you can delete the else

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,309 @@
...
+            print(f"Successfully created workspace with name: {workspace_name}.")
+            return status_code
+    elif ws_exists:  # if workspace already exists
+        print(f"Workspace already exists with name: {project}/{workspace_name}.")
+        print(f"Existing workspace details: {json.dumps(ws_exists_response.json(), indent=2)}")
+        # make user decide if they want to update/overwrite existing workspace
+        while True:  # try until user inputs valid response
+            update_existing_ws = input("Would you like to continue modifying the existing workspace? (Y/N)" + "\n")
+            if update_existing_ws.upper() in ["Y", "N"]:
+                break
```
mmorgantaylor, line 131: what's this doing?

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,309 @@
...
+
+    # capture response from API and parse out status code
+    response = requests.post(uri, headers=headers)
+    status_code = response.status_code
+
+    # print success or fail message based on status code
+    if status_code == 201:
+        print(f"Successfully created Authorization Domain with name: {auth_domain_name}.")
+        return status_code
+    elif status_code in ["403", "409"]:
```
mmorgantaylor, line 201: this can just be `if status_code in ["403", "409"]:` and can remove the `else` 3 lines down

scripts/anvil_tools/set_up_anvil_workspaces.py
schaluva: if someone puts in a value thats not an n or y (like an int or some other random letter by accident) itll keep prompting them until they put in something valid.

mmorgantaylor: ah got it. might be helpful to add a comment to say that, like `break  # don't need to keep prompting` just to help our future selves :)

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,320 @@
...
+        # return None, response.json()
+        return False, status_code, response.json()
+
+    # AD update with member success
+    print(f"Successfully updated Authorization Domain, {auth_domain_name}, with group: {email}.")
+    return True, status_code, None
+
+
+def setup_auth_domain(auth_domain_name):
+    """Create the Terra goolge group to be used as auth domain on workspaces."""
```
mmorgantaylor, line 185: can you explicitly say somewhere what the 4 values are that are returned?

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,320 @@
...
+    if status_code != 201:                           # other error (500)
+        print(f"WARNING: Failed to create Authorization Domain with name: {auth_domain_name}. Check output file for error details.")
+        return status_code, response.json(), None, None
+
+    # if AD creation success - status_code == 201
+    anvil_admin_group_name = "anvil-admins@firecloud.org"
+    is_add_mem, add_mem_status_code, add_mem_response = add_member_to_auth_domain(auth_domain_name, anvil_admin_group_name, "ADMIN")  # 204 - success otherwise fail
+
+    if not is_add_mem:           # add member to AD fail
+        return None, add_mem_response, None, False
```
mmorgantaylor, line 211: why are you returning no status code for `setup_ad_status_code` here? the group was created, but there was an error adding members, right?

mmorgantaylor: oh I think I see - the first thing returned is the success of the entire process, which failed, and you're indicating the reason for failure with the second thing returned. I think this works! It would help a lot to explicitly say what these four values represent (see previous comment)

schaluva: Oh that was because I'm using the boolean value to determine if i should print the `add_mem_response` and don't use the status_code. Should I be returning it anyway?

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,325 @@
...
+    create_ws_request["authorizationDomain"] = [{"membersGroupName": f'{auth_domain_name}'}]
+    create_ws_request["attributes"] = {}
+    # TODO: set noWorkspaceOwner = True for data delivery workspaces - picard svc is the only owner
+    create_ws_request["noWorkspaceOwner"] = False
+
+    return create_ws_request
+
+
+def add_member_to_auth_domain(auth_domain_name, email, permission):
+    """Add group with given permissions to authorization domain."""
```
mmorgantaylor, line 160: could you add returned values here too? also I think the standard is to have the final """ on its own line, so like
```
def foo():
   """Description
   inputs: whatever
   outputs: x
                  y
   """
   function stuff
```

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,325 @@
...
+    if status_code == 404:              # if workspace does not exist
+        return False, None, None
+    if status_code == 200:              # if workspace already exists
+        return True, status_code, response
+
+    return None, None, response.json()  # if other workspace exists error
+
+
+def create_workspace(workspace_name, auth_domain_name, project="anvil-datastorage"):
+    """Create the Terra workspace with given authorization domain."""
```
mmorgantaylor, line 97: also explain outputs here?

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,325 @@
...
+        print(f"WARNING: Failed to update {project}/{workspace_name} with the following user(s)/group(s): {emails}.")
+        print("Check output file for error details.")
+        return status_code, response.json(), None
+
+    print(f"Successfully updated {project}/{workspace_name} with the following user(s)/group(s): {emails}.")
+    return status_code, None, emails
+
+
+def check_workspace_exists(workspace_name, project="anvil-datastorage"):
+    """Determine if a workspace of given namespace/name already exists."""
```
mmorgantaylor, line 75: also explain outputs here :)

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,325 @@
...
+    successes = report_df.workspace_setup_status.str.count("Success").sum()
+    fails = report_df.workspace_setup_status.str.count("Failed").sum()
+    total = successes + fails
+    print(f"Number of workspaces passed set-up: {successes}/{total}")
+    print(f"Number of workspaces failed set-up: {fails}/{total}")
+    print(f"All workspace set-up (success or fail) details available in output file: {output_filename}")
+
+
+def add_members_to_workspace(workspace_name, auth_domain_name, project="anvil_datastorage"):
+    """Add members to workspace permissions."""
```
mmorgantaylor, line 43: also explain outputs here :)

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,324 @@
...
+            update_existing_ws = input("Would you like to continue modifying the existing workspace? (Y/N)" + "\n")
+            if update_existing_ws.upper() in ["Y", "N"]:
+                break
+            else:
+                print("Not a valid option. Choose: Y/N")
+        if update_existing_ws.upper() == "N":       # don't overwrite existing workspace
+            already_exists_message = f"{project}/{workspace_name} already exists. User selected not to overwrite. Try again with unique workspace name."
+            return None, already_exists_message
+
+        return True, None    # overwrite existing workspace - 200 status code for "Y"
```
mmorgantaylor: if this is what comes with overwriting existing workspace, do you want to note in the output file that this workspace already existed but the user opted to use the existing one?

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,324 @@
...
+        update_existing_ws = input("Would you like to continue modifying the existing workspace? (Y/N)" + "\n")
+        if update_existing_ws.upper() in ["Y", "N"]:
+            break
+        else:
+            print("Not a valid option. Choose: Y/N")
+    if update_existing_ws.upper() == "N":       # don't overwrite existing workspace
+        already_exists_message = f"{project}/{workspace_name} already exists. User selected not to overwrite. Try again with unique workspace name."
+        return None, already_exists_message
+
+    return True, None    # overwrite existing workspace - 200 status code for "Y"
```
mmorgantaylor, line 145: this is the case where the user said "Yes, continue modifying the existing workspace" right? In this case do you want to make note of the fact that the workspace already existed in the output doc? (you might not, that's a valid answer, just curious) - if you do want to note that, then you'd want to return that message here (and then move line 270 to before the `if not create_ws_success:` statement to write that message regardless

scripts/anvil_tools/set_up_anvil_workspaces.py
```diff
@@ -0,0 +1,324 @@
...
+    """Make the json request to pass into create_workspace()."""
+
+    # initialize empty dictionary
+    create_ws_request = {}
+
+    create_ws_request["namespace"] = project
+    create_ws_request["name"] = workspace_name
+    create_ws_request["authorizationDomain"] = [{"membersGroupName": f'{auth_domain_name}'}]
+    create_ws_request["attributes"] = {}
+    # TODO: set noWorkspaceOwner = True for data delivery workspaces - picard svc is the only owner
```
mmorgantaylor, line 158: do you want to include this as an optional input to this tool? it could default to False but you could add a flag if you wanted to set it to True?

mmorgantaylor (approved):
two small questions that you may or may not want to address, but it looks AWESOME!!!!!

PR #30, opened 2021-03-03
[DFE-419] create script to publish list of workspaces to data library in firecloud

schaluva:


scripts/anvil_tools/README.md
```diff
@@ -99,3 +99,22 @@
...
     2. `--project`, `-p`: workspace project/namespace for listed workspaces in tsv (default = anvil_datastorage)
+
+#### **publish_workspaces_to_data_library.py**
+##### Description
+    Post workspaces listed in input tsv file to the Firecloud Data Library. Must run post_workspace_attributes.py first.
+    
+    Input is a .tsv file with column:
+        1. `workspace_name`
+    Output is a .tsv file with name:
+        1. `timestamp_workspaces_published_status.tsv`
```
mmorgantaylor, line 110: could you put timestamp in brackets or something to indicate that it's not the actual word "timestamp"?

scripts/anvil_tools/publish_workspaces_to_data_library.py
```diff
@@ -0,0 +1,128 @@
...
+
+import argparse
+import datetime
+import pandas as pd
+import requests
+
+from oauth2client.client import GoogleCredentials
+
+
+def get_access_token():
```
mmorgantaylor, line 14: if this and other functions (like write_output_report) are used in multiple scripts, you should make a common scripts file and import from there so you don't have duplicate code

scripts/anvil_tools/post_workspace_attributes.py
```diff
@@ -20,53 +21,147 @@ def get_access_token():
...
+    # count success and failed workspaces and report to stdout
+    successes = workspace_status_dataframe.workspace_setup_status.str.count("Success").sum()
+    fails = workspace_status_dataframe.workspace_setup_status.str.count("Failed").sum()
+    total = successes + fails
+    print(f"Number of workspaces passed set-up: {successes}/{total}")
+    print(f"Number of workspaces failed set-up: {fails}/{total}")
+    print(f"All workspace set-up (success or fail) details available in output file: {output_filename}")
+
+
+def publish_workspace(workspace_name, project="anvil-datastorage"):
```
mmorgantaylor, line 41: i'm confused about this vs `call_publish_library_workspace_api` in the other script? are these the same?

scripts/anvil_tools/utils.py
```diff
@@ -0,0 +1,169 @@
...
+        print(response.text)
+        return False, response.text
+
+    # publishiing success
+    print(f"Successfully published {project}/{workspace_name} to Data Library.")
+    return True, response.text
+
+
+# function to add dataset/library metadata to a workspace
+def add_library_metadata_to_workspace(request, workspace_name, project="anvil-datastorage"):
```
mmorgantaylor, line 69: can you define what `request` needs to contain?

mmorgantaylor (approved):
don't want to hold you back, happy to re-review if you make significant changes!

scripts/anvil_tools/post_workspace_attributes.py
```diff
@@ -44,34 +43,60 @@ def setup_single_data_delivery_workspace(request, project="anvil-datastorage"):
...
+    workspace_dict["final_workspace_status"] = "Success"
     return workspace_dict
 
 
-def setup_data_delivery_workspaces(tsv):
-    """Create array of json requests per row in tsv file."""
+def format_dataframe(workspace_row):
+    """Format workspace row Series object and convert to json."""
+
+    list_dtype_cols = ["library:consentCodes",
```
mmorgantaylor, line 53: will this list ever change? is it all `library:` keys or just this specific subset? just thinking about whether this could ever cause unexpected behavior

scripts/anvil_tools/post_workspace_attributes.py
```diff
@@ -44,34 +43,60 @@ def setup_single_data_delivery_workspace(request, project="anvil-datastorage"):
         return workspace_dict
 
     workspace_dict["publish_workspace_status"] = "Success"
+    workspace_dict["final_workspace_status"] = "Success"
     return workspace_dict
 
 
-def setup_data_delivery_workspaces(tsv):
-    """Create array of json requests per row in tsv file."""
+def format_dataframe(workspace_row):
```
mmorgantaylor, line 50: this function name is misleading because you call it on a row, not the whole dataframe, right? maybe change the name, or have the function loop through all rows itself?

mmorgantaylor (approved):
just a question and one minor suggestion, but looks good!

scripts/anvil_tools/post_workspace_attributes.py
schaluva: its a specific subset that is defined in an attributes_definitions.json that Ruchi shared with me and she said it hasn't changed/probabaly won't change

mmorgantaylor: cool! maybe add a comment saying where the list came from and where to check for it in case of updates?

scripts/anvil_tools/post_workspace_attributes.py
```diff
@@ -94,7 +94,7 @@ def setup_data_delivery_workspaces(tsv, project="anvil-datastorage"):
         row_json_request = all_workspaces_modified_df.loc[workspace]
 
         # create validated json
-        validated_json = format_dataframe(row_json_request)
+        validated_json = format_workspace_attributes(row_json_request)
```
mmorgantaylor, line 97: beautiful!

PR #31, opened 2021-03-04
[DFE-413] allow user to pass in json instead of separate args

schaluva:
Jason mentioned that its easier for Frontline to post banners by passing in a json file instead of putting strings as individual arguments for the components of the banner. Implemented a small addition to allow for passing in a json and then posting the banner.


mmorgantaylor (approved):
looks great! assuming any needed testing looks good, lgtm!

PR #33, opened 2021-03-29
[DFE-441] modify anvil create workspaces script for van allen lab

schaluva:
1. set bucket creation to us-central1
2. remove authorization domain steps
3. modify OWNER and READER access
4. fix tsv output to have only necessary columns (take out AD cols)
5. set default project to van allen project
6. create readme and utils.py (to be refactored)

Tested by creating two workspaces - success and fail modes.

### DIFF
--- a/scripts/van_allen_tools/set_up_vanallen_workspaces.py
+++ b/scripts/van_allen_tools/set_up_vanallen_workspaces.py
@@ -109,6 +109,8 @@ def make_create_workspace_request(workspace_name, project=NAMESPACE):
     create_ws_request["name"] = workspace_name
     create_ws_request["attributes"] = {}
     create_ws_request["noWorkspaceOwner"] = False
+    # hard coded to ensure that there are no accidental egress charges across regions
+    # specific to van allen lab resources all migrating to this region
     create_ws_request["bucketLocation"] = "us-central1"
 
     return create_ws_request
--- a/scripts/van_allen_tools/set_up_vanallen_workspaces.py
+++ b/scripts/van_allen_tools/set_up_vanallen_workspaces.py
@@ -1,0 +1,189 @@
+"""Create workspaces, set up bucket in us-central region, add workspace access to users.
+
+Usage:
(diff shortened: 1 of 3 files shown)


scripts/van_allen_tools/set_up_vanallen_workspaces.py
```diff
@@ -0,0 +1,189 @@
...
+    """Make the json request to pass into create_workspace()."""
+
+    # initialize empty dictionary
+    create_ws_request = {}
+
+    create_ws_request["namespace"] = project
+    create_ws_request["name"] = workspace_name
+    create_ws_request["attributes"] = {}
+    create_ws_request["noWorkspaceOwner"] = False
+    create_ws_request["bucketLocation"] = "us-central1"
```
mmorgantaylor, line 112: this is always for sure going to be `us-central1` ? any benefit in having this be the default but allowing the user to input a different location if desired?

mmorgantaylor (approved):
one question about bucket location but otherwise lgtm!

scripts/van_allen_tools/set_up_vanallen_workspaces.py
schaluva: I thought about making it user modifiable but I think the lab wanted to be super careful and have all their resources to be in that region so as not to incur accidental egress - so i thought a hard coded region might be best! i could see how somewhere done the line at some point *maybe* they might want to change regions? what do you think? just make it a variable and set default?

mmorgantaylor: ah hm that makes sense. maybe then make it super obvious by making a top-level global variable `TARGET_BUCKET_LOCATION = "us-central1"` and then here `create_ws_request["bucketLocation"] = TARGET_BUCKET_LOCATION`

schaluva: sweet - same thought. just updated

PR #40, opened 2021-04-14
Added copy over data table and workflow, replace data table bucket

Marianie-Simeon:
- Migrate over entities 
  - Check if the entities in both workspaces match
- Migrate over workflow
- Replace data table old bucket path to new bucket path

### DIFF
--- a/scripts/van_allen_tools/migrate_van_allen_workspaces.py
+++ b/scripts/van_allen_tools/migrate_van_allen_workspaces.py
@@ -128,6 +128,150 @@ def make_create_workspace_request(workspace_name, auth_domains, namespace=NAMESP
     return create_ws_request
 
 
+def copy_workspaces_workflows(destination_workspace_namespace, destination_workspace_name, source_workspace_namespace, source_workspace_name):
+    """Copy Van Allen Lab workspaces Workflows."""
+    # Get the list of all the workflows
+    try:
+        workflow_list = fapi.list_workspace_configs(source_workspace_namespace, source_workspace_name)
+
+        # Store all the workflow names
+        workflow_name_list = []
+
+        for workflow in workflow_list.json():
+            # Get workflow config (overview config)
+            workflow_config = workflow['methodRepoMethod']
+
+            workflow_name_list.append(workflow_config['methodName'])
+
+            # Get workspace config (Detailed config with inputs, oututs, etc)
+            workspace_config = fapi.get_workspace_config(source_workspace_namespace, source_workspace_name, workflow_config['methodNamespace'], workflow_config['methodName'])
+
+            # Create a workflow based on Detailed config
+            fapi.create_workspace_config(destination_workspace_namespace, destination_workspace_name, workspace_config.json())
+            print(f"Copied {workflow_config['methodName']} workflow : over to {destination_workspace_namespace}/{destination_workspace_name}")
+
+        # Check if workflows match
+        new_workflow_list = fapi.list_workspace_configs(destination_workspace_namespace, destination_workspace_name).json()
+        if workflow_list != workflow_list:
+            print(f"Error: Data Tables don't match")
+            return False, f"Error: Data Tables don't match", None
+
+    except Exception as error:
+        return False, f"Error: {error}", None
+
+    return True, None, workflow_name_list
+
+
+def find_and_replace(attr, value, replace_this, with_this):
+    """Replace an attribute in a data table row."""
+    updated_attr = None
+    if isinstance(value, str):  # if value is just a string
+        if replace_this in value:
+            new_value = value.replace(replace_this, with_this)
+            updated_attr = fapi._attr_set(attr, new_value)
+    elif isinstance(value, dict):
+        if replace_this in str(value):
+            value_str = str(value)
+            value_str_new = value_str.replace(replace_this, with_this)
+            value_new = ast.literal_eval(value_str_new)
+            updated_attr = fapi._attr_set(attr, value_new)
+    elif isinstance(value, (bool, int, float, complex)):
+        pass
+    elif value is None:
+        pass
+    else:  # some other type, hopefully this doesn't exist
+        print('unknown type of attribute')
+        print('attr: ' + attr)
+        print('value: ' + str(value))
+
+    return updated_attr
+
+
+def update_entities(workspace_name, workspace_project, replace_this, with_this):
+    """Update Data Table"""
+    # update workspace entities
+    print("Updating DATA ENTITIES for " + workspace_name)
+
+    # get data attributes
+    response = fapi.get_entities_with_type(workspace_project, workspace_name)
+    entities = response.json()
+
+    for ent in entities:
+        ent_name = ent['name']
+        ent_type = ent['entityType']
+        ent_attrs = ent['attributes']
+        attrs_list = []
+        for attr in ent_attrs.keys():
+            value = ent_attrs[attr]
+            updated_attr = find_and_replace(attr, value, replace_this, with_this)
+            if updated_attr:
+                attrs_list.append(updated_attr)
+
+        if len(attrs_list) > 0:
+            response = fapi.update_entity(workspace_project, workspace_name, ent_type, ent_name, attrs_list)
+            if response.status_code == 200:
+                print('Updated entities:')
+                for attr in attrs_list:
+                    print('   '+str(attr['attributeName'])+' : '+str(attr['addUpdateAttribute']))
+
+
+def copy_workspace_entities(destination_workspace_namespace, destination_workspace_name, source_workspace_namespace, source_workspace_name, destination_workspace_bucket):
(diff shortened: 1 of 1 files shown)


scripts/van_allen_tools/migrate_van_allen_workspaces.py
```diff
@@ -202,49 +199,231 @@ def setup_single_workspace(workspace):
...
+
+    # Default copy_successful
+    copy_successful = True
+
+    # Default Error
+    error = "NA"
+
+    # Get the list of all the workflows
+    try:
+        workflow_list = fapi.list_workspace_configs(source_workspace_namespace, source_workspace_name)
```
schaluva, line 369: 1. also get list of workflow names from source workspace
```
source_wf_names = []
for item in json.loads(response.text):
      source_wf_names.append(item["methodRepoMethod"]["methodName"])
```
repat after the workflows are copied:
```
dest_wf_list = []
...
```

check if `source_wf_list = dest_wf_list`
if they are the same, return `True, dest_wf_list`, if they are NOT return `False, erro`

scripts/van_allen_tools/migrate_van_allen_workspaces.py
```diff
@@ -202,49 +199,231 @@ def setup_single_workspace(workspace):
...
     # per row in tsv/df
     for index, row in setup_info_df.iterrows():
 
-        row_dict = setup_single_workspace(row)
-        all_row_df = all_row_df.append(row_dict, ignore_index=True)
+        # Create Workspace
+        migration_data = setup_single_workspace(row)
+
+        # Copy over data table
+        migration_data_with_dt = copy_workspace_entities(migration_data)
```
schaluva, line 417: moving up the data table migration/copying into `setup_single_workspace` the way we did `copy_workspaces` together.

scripts/van_allen_tools/migrate_van_allen_workspaces.py
```diff
@@ -131,68 +128,214 @@ def make_create_workspace_request(workspace_name, auth_domains, namespace=NAMESP
...
+        workflow_list = fapi.list_workspace_configs(source_workspace_namespace, source_workspace_name)
+
+        # Store all the workflow names
+        workflow_name_list = []
+
+        for workflow in workflow_list.json():
+            # Get workflow config (overview config)
+            workflow_config = workflow['methodRepoMethod']
+
+            workflow_name_list.append(workflow_config['methodName'])
```
schaluva, line 144: you could do 
```workflow_name = workflow['methodRepoMethod']['methodName']
     workflow_name_list.append(workflow_name)```

I think calling it a config and then adding it to a list called workflow_name_list is a bit confusing since config is also something else

Comment (schaluva) [inline scripts/van_allen_tools/migrate_van_allen_workspaces.py:144]: maybe leave a comment that describes that `[workflow_config['methodName']` is the name of a workflow in the workspace or first defining it separately so like
```
workflow_name = [workflow_config['methodName']
workflow_name_list.append(workflow_name)
```

File: scripts/van_allen_tools/migrate_van_allen_workspaces.py [hunk 4]
```diff
@@ -131,68 +128,214 @@ def make_create_workspace_request(workspace_name, auth_domains, namespace=NAMESP
...
+            workspace_config = fapi.get_workspace_config(source_workspace_namespace, source_workspace_name, workflow_config['methodNamespace'], workflow_config['methodName'])
+
+            # Create a workflow based on Detailed config
+            fapi.create_workspace_config(destination_workspace_namespace, destination_workspace_name, workspace_config.json())
+            print(f"Copied {workflow_config['methodName']} workflow : over to {destination_workspace_namespace}/{destination_workspace_name}")
+
+        # Check if workflows match
+        new_workflow_list = fapi.list_workspace_configs(destination_workspace_namespace, destination_workspace_name).json()
+        if workflow_list != workflow_list:
+            print(f"Error: Data Tables don't match")
```
schaluva, line 156: This error says that "data tables" don't match but were looking at the workflows :)

scripts/van_allen_tools/migrate_van_allen_workspaces.py
```diff
@@ -131,68 +128,214 @@ def make_create_workspace_request(workspace_name, auth_domains, namespace=NAMESP
...
+            # Get workspace config (Detailed config with inputs, oututs, etc)
+            workspace_config = fapi.get_workspace_config(source_workspace_namespace, source_workspace_name, workflow_config['methodNamespace'], workflow_config['methodName'])
+
+            # Create a workflow based on Detailed config
+            fapi.create_workspace_config(destination_workspace_namespace, destination_workspace_name, workspace_config.json())
+            print(f"Copied {workflow_config['methodName']} workflow : over to {destination_workspace_namespace}/{destination_workspace_name}")
+
+        # Check if workflows match
+        new_workflow_list = fapi.list_workspace_configs(destination_workspace_namespace, destination_workspace_name).json()
+        if workflow_list != workflow_list:
```
schaluva, line 155: this is comparing the same list against itself - workflow_list and workflow_list

scripts/van_allen_tools/migrate_van_allen_workspaces.py
```diff
@@ -131,68 +128,214 @@ def make_create_workspace_request(workspace_name, auth_domains, namespace=NAMESP
...
 
 
+def copy_workspaces_workflows(destination_workspace_namespace, destination_workspace_name, source_workspace_namespace, source_workspace_name):
+    """Copy Van Allen Lab workspaces Workflows."""
+    # Get the list of all the workflows
+    try:
+        workflow_list = fapi.list_workspace_configs(source_workspace_namespace, source_workspace_name)
+
+        # Store all the workflow names
+        workflow_name_list = []
```
schaluva, line 138: since we make a new list below with workflow names, maybe calling this `source_workflow_names` and the one below `destination_workflow_names`?

scripts/van_allen_tools/migrate_van_allen_workspaces.py
```diff
@@ -131,68 +128,214 @@ def make_create_workspace_request(workspace_name, auth_domains, namespace=NAMESP
     return create_ws_request
 
 
+def copy_workspaces_workflows(destination_workspace_namespace, destination_workspace_name, source_workspace_namespace, source_workspace_name):
```
schaluva, line 131: the function says that it copies over multiple workspaces' workflows but it is only for one workspace

scripts/van_allen_tools/migrate_van_allen_workspaces.py
```diff
@@ -131,68 +128,214 @@ def make_create_workspace_request(workspace_name, auth_domains, namespace=NAMESP
...
+
+            # Create a workflow based on Detailed config
+            fapi.create_workspace_config(destination_workspace_namespace, destination_workspace_name, workspace_config.json())
+            print(f"Copied {workflow_config['methodName']} workflow : over to {destination_workspace_namespace}/{destination_workspace_name}")
+
+        # Check if workflows match
+        new_workflow_list = fapi.list_workspace_configs(destination_workspace_namespace, destination_workspace_name).json()
+        if workflow_list != workflow_list:
+            print(f"Error: Data Tables don't match")
+            return False, f"Error: Data Tables don't match", None
```
schaluva, line 157: maybe instead of returning 3 things we can simplify it here and do 
```
missing_workflow_names = list(set(source_workflow_names) - set(destination_workflow_names))
error = f"Not all workflow from {source_workspace_name} copied over to {destination_workspace_name}. List of missing workflows: {missing_workflow_names}"

return False, error
```

scripts/van_allen_tools/migrate_van_allen_workspaces.py
```diff
@@ -131,68 +128,214 @@ def make_create_workspace_request(workspace_name, auth_domains, namespace=NAMESP
...
+
+    except Exception as error:
+        return False, f"Error: {error}", None
+
+    return True, None, workflow_name_list
```
schaluva, line 162: Then this can be `return True, destination_workflow_names` so its a list of everything in the new workspace - which should at this point be the same as whats in the source.

scripts/van_allen_tools/migrate_van_allen_workspaces.py
```diff
@@ -131,68 +128,214 @@ def make_create_workspace_request(workspace_name, auth_domains, namespace=NAMESP
...
+    else:  # some other type, hopefully this doesn't exist
+        print('unknown type of attribute')
+        print('attr: ' + attr)
+        print('value: ' + str(value))
+
+    return updated_attr
+
+
+def update_entities(workspace_name, workspace_project, replace_this, with_this):
+    """Update Data Table"""
```
schaluva, line 191: Leaving more detail would be great! "Update data table with new paths." or something like that

schaluva (approved):

PR #42, opened 2021-04-22
Van Allen- migrate data wdl

schaluva:
current wdl being used to do test migration of data from multi region buckets to single region buckets (that have been previously made using a different script)


pipelines/migrate_bucket_regions/migrate_bucket_data_via_local.wdl
```diff
@@ -0,0 +1,138 @@
+version 1.0
+
+workflow migrate_data_via_local {
+    input {
+        String source_bucket_path
+        String destination_bucket_path
+
+        File   source_bucket_object_inventory
```
brendanreardon, line 8: Ah, I see. Yes, you wrote that the BQ table will be added as an input to the WDL. Amazing!

Marianie-Simeon (approved):

PR #43, opened 2021-04-22
[DFE-473] script to export query results to csv in gcs location

schaluva:


scripts/van_allen_tools/query_bucket_object_inventory.py
```diff
@@ -0,0 +1,120 @@
...
+import argparse
+import google.auth
+
+from google.cloud import bigquery
+
+
+# information of the destination table to be exported to a bucket as csv
+GCP_PROJECT = "dsp-fieldeng-vanallen-gcp"
+DATASET_NAME = "dfci_nih_billing_exports"
+INVENTORY_TABLE_NAME = "dfci_nih_bucket_inventory_sc_test"
```
brendanreardon, line 17: I think these lines would become 

```
GCP_PROJECT = "vanallen-gcp-nih"
DATASET_NAME = "gcs_inventory_loader"
INVENTORY_TABLE_NAME = "object_metadata"
```

brendanreardon (commented):
Does you mean that the WDL to copy data is generating a list of files to transfer to the new bucket from the gcs_object_loader BQ table? If so, we’ll have to run gcs_object_loader before running the WDL again, which is not a problem but just something to keep in mind.

PR #51, opened 2021-05-05
[DFE-481] functionality added to be able to validate DRS uris in multiple columns in data tables

schaluva:
The current DRS notebook to validate DRS sets a specific column value in which it searches for DRS URIs. However, in the case of TCGA, there are more than one column ("uuid_and_filename") that need to be searched to look for `drs://` prefixes in the row values.

To make the notebook flexible for 1 or more col names to match on, change the key structure in dictionary (`data_column_map`) to be a `list` instead of `string`. Updated the following code to be able to handle the list values.


scripts/drs_file_access_examination.ipynb
```diff
@@ -126,42 +126,57 @@
...
     "DRS_uris_dict = {}\n",
     "\n",
-    "# Going through each table in the data table and pull out the contents of the object_id column\n",
+    "# Going through each item in response (aka one row/entity in all rows from all tables)\n",
     "for table in data_tables_list:\n",
-    "    table_colums = table[\"attributes\"]\n",
-    "    object_id_column = dict(filter(lambda item: column_name_search_key in item[0], table_colums.items())) \n",
+    "    table_columns = table[\"attributes\"]  # tuples (col_name, val)\n",
+    "    # if columnns for an entity in tcga dict list, add row/entity to dictionary\n",
+    "    object_id_column = dict(filter(lambda column: column[0] in data_column_map[column_name_search_key], table_columns.items()))\n",
```
mmorgantaylor, line 163: what is this doing?

scripts/drs_file_access_examination.ipynb
```diff
@@ -126,42 +126,57 @@
...
     "    \n",
     "    # Checking if the object_id column's contents is a string and start with 'drs://' \n",
     "    if object_id_column:\n",
-    "        DRS_uri = next(iter(object_id_column.values()))\n",
-    "        if type(DRS_uri) is str and DRS_uri.startswith('drs://'):\n",
-    "            DRS_uris_dict[DRS_uri] = {'table_name': table['entityType'], \n",
+    "        for col, val in object_id_column.items():\n",
+    "            if type(val) is str and val.startswith('drs://'):\n",
```
mmorgantaylor, line 168: i think it's preferred to use `if isinstance(val, str)`

scripts/drs_file_access_examination.ipynb
```diff
@@ -59,14 +59,14 @@
...
-   "execution_count": null,
+   "execution_count": 1,
    "metadata": {
     "code_folding": []
    },
    "outputs": [],
    "source": [
     "# Set variable to either \"biodata-catalyst\", \"anvil\", \"tcga\", or \"<INSERT DRS URIs' column name>\"\n",
-    "columns_name = \"biodata-catalyst\""
+    "columns_name = \"tcga\""
```
mmorgantaylor, line 69: is this used if columns_name is anything other than `tcga` ?

scripts/drs_file_access_examination.ipynb
mmorgantaylor: where is `column_name_search_key` defined?

PR #54, opened 2021-05-13
Adding reference data and workspace data to be migrate for Van-Allen DFE-474 

Marianie-Simeon:
# Addressing 
- https://broadworkbench.atlassian.net/browse/DFE-474 
# Code
- Adding reference data and workspace data to be migrated to the new workspaces

### DIFF
--- a/scripts/van_allen_tools/migrate_van_allen_workspaces.py
+++ b/scripts/van_allen_tools/migrate_van_allen_workspaces.py
@@ -12,7 +12,7 @@ from firecloud import api as fapi
 from utils import add_tags_to_workspace, check_workspace_exists, \
     get_access_token, get_workspace_authorization_domain, \
     get_workspace_bucket, get_workspace_members, get_workspace_tags, \
-    write_output_report, write_terra_load_tsv
+    write_output_report, write_terra_load_tsv, get_workspace_attributes
 import query_bucket_object_inventory as get_source_objects_file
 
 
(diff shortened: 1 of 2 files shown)


scripts/van_allen_tools/migrate_van_allen_workspaces.py
```diff
@@ -347,6 +347,24 @@ def copy_workspace_entities(destination_workspace_namespace, destination_workspa
     return True, data_table_name_list
 
 
+def get_workspace_attributes(source_workspace_namespace, source_workspace_name):
```
schaluva, line 350: Since this function is pretty general, and can be re-used to get workspace attributes from any workspace, we could put this into the `utils.py` and just import it into this larger one! The only thing to change, and this is like a tiny minor details, would be the function variables to `project` and `workspace` - or whatever the other functions in `utils.py` use to denote a terra project and workspace name. It should be in the same directory as where this script is

scripts/van_allen_tools/migrate_van_allen_workspaces.py
```diff
@@ -347,6 +347,24 @@ def copy_workspace_entities(destination_workspace_namespace, destination_workspa
...
+    """Copy attributes from source workspace to a destination workspace."""
```
schaluva, line 351: This function only gets the workspace attributes right? It doesn't actually do any of the copying work? The description is a bit confusing since I'm expecting to see code for copying it to the destination in here.

scripts/van_allen_tools/migrate_van_allen_workspaces.py
```diff
@@ -392,8 +410,14 @@ def setup_single_workspace(workspace, ignore=[]):
     if not get_ad_success:
         return workspace_dict
 
+    # get workspace attributes including workspace data from source workspace to destination workspace
+    get_workspace_success, get_workspace_message = get_workspace_attributes(source_workspace_namespace, source_workspace_name)
```
schaluva, line 414: I know the comment describes it, however I think just the "get workspace" is something we actually do in one of the other steps to confirm if a workspace with the same name already exists. I know it would be really long as a name for a variable but could we call these `get_ws_attributes_success` and get_ws_attributes_message` or something like that so we can know for sure what step we are doing and checking?

schaluva (approved):
Thank you for taking this one on and testing it! I had some organization and just some naming variables things but otherwise it looks good to me!

PR #56, opened 2021-05-19
Change Bucket to take in a string and created a while loop to repeat pays

Marianie-Simeon:
# Addressing
- https://broadworkbench.atlassian.net/browse/DFE-506
- https://broadworkbench.atlassian.net/browse/DFE-505

# Code
- Change Bucket to take in a string
- Created a while loop to repeat requester pays

### DIFF
--- a/scripts/set_requester_pays.sh
+++ b/scripts/set_requester_pays.sh
@@ -3,7 +3,7 @@
 if (( $# < 1 )); then
   echo "Usage: $0 TERRA_PROJECT_ID PATH_TO_TERRA_BUCKET_PATH"
   echo "Unless you specify a source file or a string of buckets, the script will read buckets from a file 'buckets.txt'."
-  echo "The source file must include newline-delimited Terra bucket paths of format gs://fc-XXXXX"
+  echo "The source file must include newline-delimited Terra bucket paths"
   echo 'The string of Terra bucket paths can be formatted as "gs://fc-XXXXX gs://fc-XXXXX" or "fc-XXXXX fc-XXXXX"'
   echo "i.e you can run `./set_requester_pays.sh project_id fc-12345`"
   echo "NOTE: this script requires you to be authed as your firecloud.org admin account."
@@ -19,7 +19,7 @@ if (( $# < 1 )); then
     fi
   # If there is more that 2 arguments, check if the bucket list has qutoes around it
   elif (( $# > 2 )); then
-    echo 'The Terra bucket paths must be formatted as "gs://fc-XXXXX gs://fc-XXXXX gs://fc-XXXXX"'
+    echo 'The Terra bucket paths can be formatted as "gs://fc-XXXXX gs://fc-XXXXX" or "fc-XXXXX fc-XXXXX"'
     echo "NOTE: this script requires you to be authed as your firecloud.org admin account."
     exit 0
 fi
--- a/scripts/set_requester_pays.sh
+++ b/scripts/set_requester_pays.sh
@@ -4,7 +4,8 @@ if (( $# < 1 )); then
   echo "Usage: $0 TERRA_PROJECT_ID PATH_TO_TERRA_BUCKET_PATH"
   echo "Unless you specify a source file or a string of buckets, the script will read buckets from a file 'buckets.txt'."
   echo "The source file must include newline-delimited Terra bucket paths of format gs://fc-XXXXX"
-  echo 'The string of Terra bucket paths must be formatted as "gs://fc-XXXXX gs://fc-XXXXX gs://fc-XXXXX"'
+  echo 'The string of Terra bucket paths can be formatted as "gs://fc-XXXXX gs://fc-XXXXX" or "fc-XXXXX fc-XXXXX"'
+  echo "i.e you can run `./set_requester_pays.sh project_id fc-12345`"
   echo "NOTE: this script requires you to be authed as your firecloud.org admin account."
   exit 0
   elif (( $# == 1 )); then
@@ -41,6 +42,8 @@ echo "Gatorcounting for 10 seconds while iam change goes into effect"
 echo ""
 echo "NOTE: if you get an error message saying AccessDeniedExeption: 403"
 echo "THEN don't worry, just wait until it shows 'Enabling requester pays."
+echo ""
+echo "retrying in 10 seconds"
 echo ""
 sleep 10
 
--- a/scripts/set_requester_pays.sh
+++ b/scripts/set_requester_pays.sh
@@ -47,12 +47,18 @@ sleep 10
 COUNTER=0
 
 for BUCKET in $BUCKETS; do
-  while ! gsutil requesterpays set on ${BUCKET}
+  if [[ $BUCKET == *"gs://"* ]]; then
+      BUCKET_PATH=$BUCKET
+    else
+      BUCKET_PATH="gs://$BUCKET"
+  fi
+  while ! gsutil requesterpays set on ${BUCKET_PATH}
     do
-      sleep 10
       let COUNTER=COUNTER+1
-      if (( $COUNTER > 20 )); then
-        echo "Error - requesterpays Timeout"
+      echo "retrying in 10 seconds - attempt ${COUNTER}/6"
+      sleep 10
+      if (( $COUNTER > 6 )); then
+        echo "Error - Requesterpays Timeout"
         exit 0
       fi
     done
--- a/scripts/set_requester_pays.sh
+++ b/scripts/set_requester_pays.sh
@@ -1,18 +1,14 @@
 #!/bin/bash
 
-if (( $# < 1 )); then
-  echo "Usage: $0 TERRA_PROJECT_ID [PATH_TO_TERRA_BUCKET_LIST_FILE]"
-  echo "Unless you specify a source file, the script will read buckets from a file 'buckets.txt'."
-  echo "The source file must include newline-delimited Terra bucket paths of format gs://fc-XXXXX"
+if (( $# < 2 )); then
+  echo "Usage: $0 TERRA_PROJECT_ID PATH_TO_TERRA_BUCKET_PATH"
+  echo "The Terra bucket paths must be formatted as gs://fc-XXXXX"
   echo "NOTE: this script requires you to be authed as your firecloud.org admin account."
   exit 0
-elif (( $# == 1 )); then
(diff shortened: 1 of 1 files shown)


scripts/set_requester_pays.sh
```diff
@@ -1,18 +1,14 @@
...
   echo "NOTE: this script requires you to be authed as your firecloud.org admin account."
   exit 0
-elif (( $# == 1 )); then
-  BUCKETS=$(cat buckets.txt)
-elif (( $# == 2 )); then
-  BUCKETS=$(cat $2)
 fi
 
 PROJECT_ID=$1
+BUCKETS=$2
```
schaluva, line 11: since before we had passed in a newline delimited .txt file to list the buckets in question, what format should the buckets be passed in as now? or are we just saying that we pass in a string of bucket ids?

scripts/set_requester_pays.sh
```diff
@@ -32,13 +28,15 @@ echo "Gatorcounting for 10 seconds while iam change goes into effect"
 echo ""
 echo "NOTE: if you get an error message saying:"
```
mmorgantaylor, line 29: this isn't part of your PR, but after using this a bunch, I find the following long error message example confusing - could we remove the next line and change this one to be just 
`echo "NOTE: if you get an error message saying AccessDeniedExeption: 403"`

scripts/set_requester_pays.sh
```diff
@@ -32,13 +28,15 @@ echo "Gatorcounting for 10 seconds while iam change goes into effect"
...
 echo "    AccessDeniedException: 403 ${USER_EMAIL} does not have storage.buckets.update access to the Google Cloud Storage bucket."
-echo "THEN gatorcount 10 more seconds and run this again."
+echo "THEN don't worry, just wait until it shows 'Enabling requester pays."
 echo ""
 sleep 10
 
 for BUCKET in $BUCKETS; do
-  # set requester pays
-  gsutil requesterpays set on ${BUCKET} || exit 1
+  while ! gsutil requesterpays set on ${BUCKET}
```
mmorgantaylor, line 36: what happens if we get a non-403 error? can we add a timeout in case something else is wrong, so this doesn't go on forever if there's something else wrong besides a permissions issue?

scripts/set_requester_pays.sh
mmorgantaylor: Ideally we'd maintain both functionalities

mmorgantaylor (changes requested):
some comments - most major is to maintain both methods of providing a bucket or list of buckets (just the string for a single one, or a text file containing multiple)

scripts/set_requester_pays.sh
mmorgantaylor: your changes resolve the timeout issue, but what about a non-403 error?

scripts/set_requester_pays.sh
mmorgantaylor: this is one solution, and you could argue it's the best solution - but what do we do in the case where someone hands us a text file of 50 newline-delimited buckets? we have to go in and reformat it to be separated by spaces? perhaps we could in that case have a small script that will format a list of newline-delimited bucket strings into the input needed for this function?

Marianie-Simeon: I can test it, but I think it would work if you pass in a text file instead

scripts/set_requester_pays.sh
```diff
@@ -23,22 +32,30 @@ ROLE="organizations/${ORG_ID}/roles/RequesterPaysToggler"
...
 # # if needed for troubleshooting, this command retrieves the existing policy
 # gcloud beta projects get-iam-policy $PROJECT_ID | grep -A 1 -B 1 "${MEMBER}"
 
 echo ""
 echo "Gatorcounting for 10 seconds while iam change goes into effect"
 echo ""
-echo "NOTE: if you get an error message saying:"
-echo "    AccessDeniedException: 403 ${USER_EMAIL} does not have storage.buckets.update access to the Google Cloud Storage bucket."
-echo "THEN gatorcount 10 more seconds and run this again."
+echo "NOTE: if you get an error message saying AccessDeniedExeption: 403"
```
mmorgantaylor, line 42: I think that if we add an `echo "retrying in 10 seconds"` before the sleep 10 in line 52, then we can remove this and the next line altogether

scripts/set_requester_pays.sh
mmorgantaylor: Marianie confirmed that the error will be printed, so if it's not a 403, the user will know.

scripts/set_requester_pays.sh
```diff
@@ -1,15 +1,24 @@
 #!/bin/bash
 
 if (( $# < 1 )); then
-  echo "Usage: $0 TERRA_PROJECT_ID [PATH_TO_TERRA_BUCKET_LIST_FILE]"
-  echo "Unless you specify a source file, the script will read buckets from a file 'buckets.txt'."
+  echo "Usage: $0 TERRA_PROJECT_ID PATH_TO_TERRA_BUCKET_PATH"
+  echo "Unless you specify a source file or a string of buckets, the script will read buckets from a file 'buckets.txt'."
   echo "The source file must include newline-delimited Terra bucket paths of format gs://fc-XXXXX"
+  echo 'The string of Terra bucket paths must be formatted as "gs://fc-XXXXX gs://fc-XXXXX gs://fc-XXXXX"'
```
mmorgantaylor, line 7: the ONLY other thing I can think that would make this even easier on us, is don't require the `gs://` prefix, and if it's not there, put it there in the code
i.e. so i can run this just with `./set_requester_pays.sh project_id fc-12345`

scripts/set_requester_pays.sh
```diff
@@ -23,22 +32,30 @@ ROLE="organizations/${ORG_ID}/roles/RequesterPaysToggler"
...
+
 for BUCKET in $BUCKETS; do
-  # set requester pays
-  gsutil requesterpays set on ${BUCKET} || exit 1
+  while ! gsutil requesterpays set on ${BUCKET}
+    do
+      sleep 10
+      let COUNTER=COUNTER+1
+      if (( $COUNTER > 20 )); then
+        echo "Error - requesterpays Timeout"
```
mmorgantaylor, line 55: this is minor, but I think a more descriptive message here would be something like "Errors not resolved with retries - please check the error message to resolve the problem"

scripts/set_requester_pays.sh
```diff
@@ -23,22 +33,38 @@ ROLE="organizations/${ORG_ID}/roles/RequesterPaysToggler"
...
+echo "retrying in 10 seconds"
 echo ""
 sleep 10
 
+COUNTER=0
+
 for BUCKET in $BUCKETS; do
-  # set requester pays
-  gsutil requesterpays set on ${BUCKET} || exit 1
+  if [[ $BUCKET == *"gs://"* ]]; then
```
mmorgantaylor, line 53: can we remove the first * here? if the `gs://` is somehow in the middle of the string, we have bigger problems, but we can be reasonably confident that it should be only at the beginning

scripts/set_requester_pays.sh
```diff
@@ -23,22 +33,38 @@ ROLE="organizations/${ORG_ID}/roles/RequesterPaysToggler"
...
+      BUCKET_PATH=$BUCKET
+    else
+      BUCKET_PATH="gs://$BUCKET"
+  fi
+  while ! gsutil requesterpays set on ${BUCKET_PATH}
+    do
+      let COUNTER=COUNTER+1
+      echo "retrying in 10 seconds - attempt ${COUNTER}/6"
+      sleep 10
```
mmorgantaylor, line 62: we should move this sleep to after the possible timeout - no need to wait 10 more seconds if we're out of retries

mmorgantaylor: also the echo should be after the possible timeout - we don't want to say we're retrying if we're about to error out on a timeout

scripts/set_requester_pays.sh
```diff
@@ -1,15 +1,25 @@
...
+  elif (( $# == 1 )); then
+    BUCKETS=$(cat buckets.txt)
+  elif (( $# == 2 )); then
+    if [[ $2 == *"."* ]]; then
+      BUCKETS=$(cat $2)
+    else
+      BUCKETS=$2
+    fi
+  elif (( $# > 2 )); then
+    echo 'The Terra bucket paths must be formatted as "gs://fc-XXXXX gs://fc-XXXXX gs://fc-XXXXX"'
```
mmorgantaylor, line 20: this requirement is no longer true, right? we can tolerate a lack of `gs://` prefix now

scripts/set_requester_pays.sh
```diff
@@ -1,15 +1,25 @@
...
-  BUCKETS=$(cat $2)
+  elif (( $# == 1 )); then
+    BUCKETS=$(cat buckets.txt)
+  elif (( $# == 2 )); then
+    if [[ $2 == *"."* ]]; then
+      BUCKETS=$(cat $2)
+    else
+      BUCKETS=$2
+    fi
+  elif (( $# > 2 )); then
```
mmorgantaylor, line 19: if we take in multiple buckets inline, won't $# be > 2? like if you run `./set_requester_pays.sh project_id bucket1 bucket2 bucket3` then $# will be 4, right?

scripts/set_requester_pays.sh
```diff
@@ -23,22 +34,38 @@ ROLE="organizations/${ORG_ID}/roles/RequesterPaysToggler"
...
 echo ""
 echo "Gatorcounting for 10 seconds while iam change goes into effect"
 echo ""
-echo "NOTE: if you get an error message saying:"
-echo "    AccessDeniedException: 403 ${USER_EMAIL} does not have storage.buckets.update access to the Google Cloud Storage bucket."
-echo "THEN gatorcount 10 more seconds and run this again."
+echo "NOTE: if you get an error message saying AccessDeniedExeption: 403"
+echo "THEN don't worry, just wait until it shows 'Enabling requester pays."
+echo ""
+echo "retrying in 10 seconds"
```
mmorgantaylor, line 47: change to "waiting 10 seconds before first attempt"

mmorgantaylor (approved):
this looks awesome! if you've tested this final version in all its different ways it can be called (one bucket, multiple buckets inline, file of newline-delimited buckets) and it works, then merge away!!

PR #57, opened 2021-05-27
[DFE-512] bash script to check if files exist at GCS location

schaluva:


scripts/anvil_tools/file_exists_checker.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+set -e
+
```
mmorgantaylor, line 3: could you add a "usage" section? that checks if you don't give it the needed arguments, it prints out how to call the script? see the set_requester_pays.sh script for an example

scripts/anvil_tools/file_exists_checker.sh
```diff
@@ -0,0 +1,36 @@
...
+# define variables for cmd parameters
+excel_file=$1
+check_col_num=$2
+# get basename for excel file
+excel_basename="$(basename "$excel_file" .xlsx)"
+
+# convert xlsx to txt
+echo "Converting xlsx to tsv."
+ssconvert "$excel_file" "${excel_basename}.txt"
+# substitute comma to tab | remove header | get col of crai paths > write tmp file
```
mmorgantaylor, line 13: i love how you do these comments with the pipes - it's so clear 😍

scripts/anvil_tools/file_exists_checker.sh
```diff
@@ -0,0 +1,36 @@
...
+tr "," "\t" < "${excel_basename}.txt" | sed -e 1d | cut -f"$check_col_num" > "${excel_basename}.tmp"
+
+echo "Checking if files exist."
+# create tmp file for col of exists or not exists values
+echo "file_exists" >> exists_file.tmp
```
mmorgantaylor, line 18: did you mean to include this echo? what is it for?

mmorgantaylor: oh is it the header for the column that's added later? a comment to that effect would be 💯

scripts/anvil_tools/file_exists_checker.sh
```diff
@@ -0,0 +1,36 @@
...
+
+# for each path in file of paths
+while IFS="" read -r gs_path
+    do
+        if ! gsutil -q stat "$gs_path"
+            then
+                echo "File does not exist." >> exists_file.tmp
```
mmorgantaylor, line 25: perhaps make these "true" and "false" since they're in the column named "file_exists" ?

scripts/anvil_tools/file_exists_checker.sh
```diff
@@ -0,0 +1,36 @@
...
+        else
+                echo "File exists." >> exists_file.tmp
+        fi
+done < "${excel_basename}.tmp"
+
+echo "Creating new excel file with additional column - 'file_exists'."
```
mmorgantaylor, line 31: add the name of the new file to this echo, or echo it later so the user can easily find it?

mmorgantaylor (approved):
looks great, some minor comments/suggestions for usability

scripts/anvil_tools/README.md
```diff
@@ -44,7 +44,7 @@
     Locally
         `./file_exists_checker EXCEL_FILE COLUMN_NUMBER_TO_CHECK`
     Docker
-        `docker run --rm -it -v "$HOME"/.config:/.config -v "$HOME"/local_data_directory/:/data broadinstitute/horsefish:latest bash -c "cd data; ./file_exists_checker EXCEL_FILE COLUMN_NUMBER_TO_CHECK"`
+        `docker run --rm -it -v "$HOME"/.config:/.config -v "$HOME"/local_data_directory/:/data broadinstitute/horsefish:latest bash -c "export CLOUDSDK_PYTHON=python2.7; cd data; ./../scripts/anvil_tools/file_exists_checker.sh EXCEL_FILE COLUMN_NUMBER_TO_CHECK; unset CLOUDSDK_PYTHON"`
```
mmorgantaylor: noooooo python 2 nooooooo - but seriously, can we please please please not use python 2?

mmorgantaylor: oh is this outdated now?

scripts/anvil_tools/README.md
```diff
@@ -24,13 +24,33 @@
...
+    Input is:
+        1. excel file with column of gs://XXXX file paths to validate (must end in .xlsx)
+        2. number of column with gs://XXXX file paths (must be an integer value)
+    Output is a .xlsx file with name:
+        1. `{input_excel_filenamme}_with_exists.xlsx`
+##### Usage
+    Locally
+        `./file_exists_checker EXCEL_FILE COLUMN_NUMBER_TO_CHECK`
+    Docker
+        `docker run --rm -it -v "$HOME"/.config:/.config -v "$HOME"/local_data_directory/:/data broadinstitute/horsefish bash -c "export CLOUDSDK_PYTHON=python2.7; cd data; ./../scripts/anvil_tools/file_exists_checker.sh EXCEL_FILE COLUMN_NUMBER_TO_CHECK; unset CLOUDSDK_PYTHON"`
```
mmorgantaylor, line 47: wait so why do we need to use python2 here?

scripts/anvil_tools/file_exists_checker.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+set -e
+
+# if num arguments != 2 || file does not exist || filename suffix != ".xlsx" || col number != integer
+if [ "$#" -ne 2 ] || [ ! -f "$1" ] || grep -q xlsx "$1" || [[ ! $2 =~ ^[[:digit:]]+$ ]];
+  then
+    echo "Missing required arguments or incorrect argument format/type."
+    echo "Usage: $0 [EXCEL_FILE] [COLUMN_NUMBER_TO_CHECK]"
+    echo "EXCEL_FILE: path to input excel (.xlsx) file."
+    echo "COLUMN_NUMBER_TO_CHECK: integer/number of column with gs://fc-XXXX file paths validate."
```
mmorgantaylor, line 10: 1-indexed, i assume?

scripts/anvil_tools/README.md
schaluva: when running the script, looks like cloudsdk is using python3.9 which is giving an error about one of the imported packages it uses. the internet says to use python3.7 specifically so i just set the env variable for this script.

schaluva: `AttributeError: module 'base64' has no attribute 'decodestring'` specific error when python is 3.9

scripts/anvil_tools/file_exists_checker.sh
schaluva: not really, with bash the script name is also a "parameter, aka $0

scripts/anvil_tools/README.md
mmorgantaylor: wait so did you mean to have this be `3.7` instead of `2.7` ?

scripts/anvil_tools/file_exists_checker.sh
mmorgantaylor: no i mean... if the column to check is the second column, i should have this input param be 2, not 1, right?

scripts/anvil_tools/README.md
schaluva: yep it should be 3.7 now (i dont know if it refreshed in git)

scripts/anvil_tools/file_exists_checker.sh
schaluva: yes! the column number in the excel is 1 index.

PR #59, opened 2021-06-16
DFE-[493] create standalone modified script to update workspace data models to compact DRS identifiers

schaluva:
this PR is the exact same code as the notebook version except that I:
took out the code that write the tsv files to the workspace bucket and replace with code to write the tsv files to the local machine. this change was made so that we don't leave behind a notebook and the tsv files as artifacts in the TCGA data workspaces at the end of the update - this seemed safer than going through and running deletes on TCGA workspaces.

script was tested on clones of the tcga workspaces.


scripts/drs_v1.1_uri_migration.py
```diff
@@ -0,0 +1,112 @@
...
+        with open(updated_tsv_name, 'w') as outfile:
+            # get keys from OrderedDictionary and write rows, separate with tabs
+            writer = csv.DictWriter(outfile, tsv_headers, delimiter="\t")
+            writer.writeheader()
+            writer.writerows(drs_dict_table)
+
+        # list of the columns that are scoped to be updated if re-run with dry_run = False
+        modified_cols = list(modified_cols)
+        if dry_run:
+            print(f"Columns in the {etype} table that *will be* be updated when notebook is re-run with `dry_run = False`:")
```
mmorgantaylor, line 63: here this should be `when this script is re-run without the --dry-run flag` right?

mmorgantaylor (approved):
looks great! one teeny tiny documentation change

scripts/drs_v1.1_uri_migration.py
schaluva: ah yes the instructions in there are for the notebook. good catch.

PR #61, opened 2021-06-23
[DFE-524] split large tsv file by workspace and project columns and push subset tsv to terra data model

schaluva:
Description:
AnVIL collaborators will submit a single tsv file containing data they want uploaded as tables to workspaces. There will be two columns labeled as `workspace_name` and `workspace_project` PER row - multiple rows can have the same combination of workspace and project information. This just means that the entire set of rows will be uploaded to the same workspace.

The script parses on the unique combination of `workspace_name` and `workspace_project` --> subsets the large tsv file into smaller tsv files, representative of a single workspace's data table contents --> calls  functions from `batch_upsert_entities_standard.py` to create the json request PER subsetted tsv --> loads the table to its respective workspace.

Tests:
1. `convert_string_to_array_unit_test.py` will check that if array type columns/attributes were provided, that they are formatted in the correct way so that Terra will accept and represent them correctly as "# items" in the UI. The unit test imports the test function directly from `batch_upsert_entities_standard.py`.
2. Dockerhub tested from this feature branch to make sure that the README commands to run locally as well as from docker work as expected.

Notes:
1. The input tsv file will have the same formatted columns/attributes for ALL the workspaces. This means that this script can only create a single entity table. Every workspace will have a table named `entity_table_name` with its pertinent rows.
2. The script can handle columns/attributes that are intended to be arrays - for this the user has to provide an additional file input - it is optional and defaults to all columns/attributes being of NON-array type.
3. There is already a script named `batch_upsert_entities.py` that was created for covid-sabeti lab work but since the work around that is a bit up in the air, I did not want to modify the existing script. It is linked to some workflows that may or may not be in flight. Therefore, I made a new version, with the suffix `standard` to denote that this version can be used for every other case where you want to upload a .tsv to a workspace. Once sabeti lab work is confirmed to no longer depend on the specific version of the script that already exists, we can remove it.

### DIFF
--- a/scripts/anvil_tools/batch_upsert_entities_standard.py
+++ b/scripts/anvil_tools/batch_upsert_entities_standard.py
@@ -38,10 +38,10 @@ def call_rawls_batch_upsert(workspace_name, project, request):
     print(f"Successfully uploaded entities." + "\n")
 
 
-def write_request_json(request):
+def write_request_json(request, filename_prefix):
     """Create output file with json request."""
 
-    save_name = "batch_upsert_request.json"
+    save_name = f"{filename_prefix}_batch_upsert_request.json"
     with open(save_name, "w") as f:
         f.write(request)
 
@@ -58,7 +58,7 @@ def convert_string_to_list(input_string):
 def create_upsert_request(tsv, array_attr_cols=None):
     """Generate the request body for batchUpsert API."""
 
-    # check if the tsv is formatted correctly - exit script if not in right load format
+    # check tsv format: data model load tsv requirement "entity:table_name_id" or "membership:table_name_id" -> else exit
     entity_type_col_name = tsv.columns[0]                               # entity:entity_name_id
     entity_type = entity_type_col_name.rsplit("_", 1)[0].split(":")[1]  # entity_name
 
@@ -67,13 +67,16 @@ def create_upsert_request(tsv, array_attr_cols=None):
         return
 
     # replace the "entity:col_name_id" with just "col_name" in df
-    # if not replaced, each "attributeName" in the template_make_single_attr becomes "entity:entity_name_id"entity_type_col_name
+    # if not replaced, "attributeName" in the template_make_single_attr becomes "entity:entity_name_id" instead of just entity_name
     # when the API request is made, its read as multiple columns with the "entity" prefix which is illegal
+    # this is specific just to the first column where the format is required for terra load tsv files
     tsv.rename(columns={entity_type_col_name: entity_type}, inplace=True)
 
-    # templates for request body components
+    # templates
+    # for request body skeleton
     template_req_body = """{"name":"VAR_ENTITY_ID", "entityType":"VAR_ENTITY_TYPE", "operations":[OPERATIONS_LIST]}"""
 
+    # for operations - create/update: list attribute, add list member to attribute if list, single attribute and value
     template_make_list_attr = '{"op":"CreateAttributeValueList","attributeName":"VAR_ATTRIBUTE_LIST_NAME"},'
     template_add_list_member = '{"op":"AddListMember","attributeListName":"VAR_ATTRIBUTE_LIST_NAME", "newMember":"VAR_LIST_MEMBER"},'
     template_make_single_attr = '{"op":"AddUpdateAttribute","attributeName":"VAR_ATTRIBUTE_NAME", "addUpdateAttribute":"VAR_ATTRIBUTE_MEMBER"},'
@@ -86,23 +89,25 @@ def create_upsert_request(tsv, array_attr_cols=None):
         # get the entity_id (row id - must be unique)
         entity_id = str(row[0])
 
-        # if there are array attribute columns
+        # if array columns/attributes
         if array_attr_cols:
-            # for each column that is an array
+            # for each array column
             for col in array_attr_cols:
+                # add column name to template for a single attribute
                 single_attribute_request += template_make_list_attr.replace("VAR_ATTRIBUTE_LIST_NAME", col)
-                # convert given value from tsv -> translate to a string -> back into an array
+                # convert string value -> back into an array: [foo,bar] (str) --> ['foo', 'bar'] (list)
                 attr_values = convert_string_to_list(row[col])
                 # for each item in array, add row attribute value to the above created list attribute
                 for val in attr_values:
                     single_attribute_request += template_add_list_member.replace("VAR_LIST_MEMBER", val).replace("VAR_ATTRIBUTE_LIST_NAME", col)
 
-            # set the column values for the single attribute list based on which of the full tsv columns are array attributes
+            # get single attribute list based on which of the full tsv columns are array attributes
             single_attr_cols = list(set(list(tsv.columns)) - set(array_attr_cols))
+        # if no array columns/attributes
         else:
             single_attr_cols = list(tsv.columns)
 
-        # if there are single attribute columns
+        # if there are single attribute columns - cases where all columns are arrays, single_attr_cols would be empty
         if single_attr_cols:
             # for each column that is not an array
             for col in single_attr_cols:
@@ -119,13 +124,12 @@ def create_upsert_request(tsv, array_attr_cols=None):
         # put operations list into request body template
         final_single_attribute_request = final_single_attribute_request.replace("OPERATIONS_LIST", single_attribute_request)
 
+        # add the single workspace request to largest list of individual workspace requests
         all_attributes_request.append(final_single_attribute_request)
 
     # remove quotes around elements of the list but keep the brackets - using sep() or join() get rid of the brackets
+    # ['{}', '{}', '{}'] (api fails) --> [{}, {}, {}] (api succeeds)
     all_attributes_request_formatted = '[%s]' % ','.join(all_attributes_request)
-
-    # write out a json of the request body
-    write_request_json(all_attributes_request_formatted)
 
     return all_attributes_request_formatted
 
--- a/scripts/anvil_tools/batch_upsert_entities_standard.py
+++ b/scripts/anvil_tools/batch_upsert_entities_standard.py
@@ -1,5 +1,4 @@
 import argparse
-import pandas as pd
 import requests
 
 from oauth2client.client import GoogleCredentials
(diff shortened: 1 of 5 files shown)


scripts/anvil_tools/batch_upsert_entities_standard.py
```diff
@@ -0,0 +1,148 @@
...
+    # remove single & double quotes, remove spaces, remove [ ], separate remaining string on commas (resulting in a list)
+    output_list = str(input_string).replace("'", '').replace('"', '').replace(" ", "").strip('[]').split(",")
+
+    return output_list
+
+
+def create_upsert_request(tsv, array_attr_cols=None):
+    """Generate the request body for batchUpsert API."""
+
+    # check tsv format: data model load tsv requirement "entity:table_name_id" or "membership:table_name_id" -> else exit
```
mmorgantaylor, line 61: can you add this requirement to the README for input for this function? it only has the workspace/project requirements as far as I can tell

scripts/anvil_tools/batch_upsert_entities_standard.py
```diff
@@ -0,0 +1,148 @@
...
+        single_attribute_request = ''''''
+        # get the entity_id (row id - must be unique)
+        entity_id = str(row[0])
+
+        # if array columns/attributes
+        if array_attr_cols:
+            # for each array column
+            for col in array_attr_cols:
+                # add column name to template for a single attribute
+                single_attribute_request += template_make_list_attr.replace("VAR_ATTRIBUTE_LIST_NAME", col)
```
mmorgantaylor, line 97: this is a creative way to do this - you don't need to change it, but a more functional programming way to do it would have been to have a function for each operation like
```
def get_make_list_attr_request(var_attribute_list_name):
    return '{"op":"CreateAttributeValueList","attributeName":"' + var_attribute_list_name + '"},'

def get_add_list_member_request(var_attribute_list_name, var_list_member):
    return '{"op":"AddListMember","attributeListName":"' + var_attribute_list_name + '", "newMember":"' + var_list_member + '"},'
```
etc and one for make_single_attr as well.
to me that would be easier to parse than the replace string method. just food for thought!

scripts/anvil_tools/batch_upsert_entities_standard.py
```diff
@@ -0,0 +1,148 @@
...
+                # add the request for attribute to list
+                single_attribute_request += template_make_single_attr.replace("VAR_ATTRIBUTE_MEMBER", attr_value).replace("VAR_ATTRIBUTE_NAME", col)
+
+        # remove trailing comma from the last request template
+        single_attribute_request = single_attribute_request[:-1]
+
+        # put entity_type and entity_id in request body template
+        final_single_attribute_request = template_req_body.replace("VAR_ENTITY_ID", entity_id).replace("VAR_ENTITY_TYPE", entity_type)
+        # put operations list into request body template
+        final_single_attribute_request = final_single_attribute_request.replace("OPERATIONS_LIST", single_attribute_request)
```
mmorgantaylor, line 125: this and the previous logic is pretty hard to follow - I would slightly more strongly recommend the functional approach I outlined above, if you're up for it. I think it'll make our future selves happier...

scripts/anvil_tools/batch_upsert_entities_standard.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # add the single workspace request to largest list of individual workspace requests
+        all_attributes_request.append(final_single_attribute_request)
+
+    # remove quotes around elements of the list but keep the brackets - using sep() or join() get rid of the brackets
+    # ['{}', '{}', '{}'] (api fails) --> [{}, {}, {}] (api succeeds)
+    all_attributes_request_formatted = '[%s]' % ','.join(all_attributes_request)
```
mmorgantaylor, line 132: does json.dumps achieve this for you?

mmorgantaylor: possibly in combination with making the requests actually dictionaries rather than strings...

scripts/anvil_tools/split_and_push_data_model_tsvs.py
```diff
@@ -0,0 +1,66 @@
...
+        # get rows that match the combination of workspace_name and workspace_project, drop the workspace identifying columns (name + project)
+        workspace_tsv = tsv_all.loc[(tsv_all['workspace_name'] == workspace_name) & (tsv_all['workspace_project'] == workspace_project)] \
+            .drop(['workspace_name', 'workspace_project'], axis=1)
+
+        # pass in workspace subsetted tsv to convert the df into properly formatted batchUpsert request
+        print(f"Starting entity updates to {workspace_project}/{workspace_name}:")
+        print(f"Creating json request.")
+        json_request = batch_upsert_entities_standard.create_upsert_request(workspace_tsv, array_attr_cols)
+
+        # write out a json of the request body
```
mmorgantaylor, line 42: why do you need to write this local file? is this just for troubleshooting? can we make this section optional with a flag if so? (always prefer to avoid creating unneeded files)

mmorgantaylor (approved):
this looks fine - I think some minor refactoring would make it a lot clearer, but I'll leave it to you whether it's worth the time to do it

scripts/anvil_tools/batch_upsert_entities_standard.py
schaluva: It was listed here https://github.com/broadinstitute/horsefish/pull/61/files#diff-19819e317aedad2e9f89b9364f03af1b9ea7bc27949dd41866e299a5d435a37aR151

but i almost missed it myself so i tried to make it more clear to the eye!

scripts/anvil_tools/batch_upsert_entities_standard.py
schaluva: great idea - it does make it ALOT clearer

scripts/anvil_tools/batch_upsert_entities_standard.py
schaluva: i think this is more cleaned up and organized with the above suggestion implemented. i also edited variable names and comments to more clearly explain whats happening at each step.

scripts/anvil_tools/split_and_push_data_model_tsvs.py
schaluva: I thought it might be useful to have if you needed it for provenance or something of the sort but agreed that it should just be optional

scripts/anvil_tools/batch_upsert_entities_standard.py
schaluva: I think something about this API that I can't seem to figure out how to operate without the string format. I left this part as is for the moment but will revisit at a later time!

scripts/anvil_tools/batch_upsert_entities_standard.py
```diff
@@ -55,6 +55,30 @@ def convert_string_to_list(input_string):
...
+    return '{"op":"AddListMember","attributeListName":"' + var_attribute_list_name + '", "newMember":"' + var_attribute_list_member + '"},'
+
+
+def create_non_array_attr_operation(var_attribute_name, var_attribute_value):
+    """Return json request for a single operation to create a non-array attribute."""
+
+    return '{"op":"AddUpdateAttribute","attributeName":"' + var_attribute_name + '", "addUpdateAttribute":"' + var_attribute_value + '"},'
+
+
+def create_single_entity_request(var_entity_id, var_entity_type, single_entity_operations_list):
```
mmorgantaylor, line 76: is the list input an actual list, or a string that looks like a list? if it's an actual list object, i'm not sure how that will behave with this string interpolation? if it's not a list object, i'd clarify what this input should look like in the function description comment

scripts/anvil_tools/batch_upsert_entities_standard.py
```diff
@@ -55,6 +55,30 @@ def convert_string_to_list(input_string):
...
+def add_list_member_operation(var_attribute_list_name, var_attribute_list_member):
+    """Return json request for a single operation to add a list member to an attribute of type array/list."""
+
+    return '{"op":"AddListMember","attributeListName":"' + var_attribute_list_name + '", "newMember":"' + var_attribute_list_member + '"},'
+
+
+def create_non_array_attr_operation(var_attribute_name, var_attribute_value):
+    """Return json request for a single operation to create a non-array attribute."""
+
+    return '{"op":"AddUpdateAttribute","attributeName":"' + var_attribute_name + '", "addUpdateAttribute":"' + var_attribute_value + '"},'
```
mmorgantaylor, line 73: why not remove the comma here so you don't have to do it later? actually I'd remove the comma in all of these functions and _add_ the commas when you're making the composite list

scripts/anvil_tools/batch_upsert_entities_standard.py
```diff
@@ -72,66 +96,58 @@ def create_upsert_request(tsv, array_attr_cols=None):
...
                 # get value in col from df
                 attr_value = str(row[col])
-                # add the request for attribute to list
-                single_attribute_request += template_make_single_attr.replace("VAR_ATTRIBUTE_MEMBER", attr_value).replace("VAR_ATTRIBUTE_NAME", col)
+                # get operation json with the row (entity) value to the non-array column (attribute)
+                single_entity_operations += create_non_array_attr_operation(col, attr_value)
 
         # remove trailing comma from the last request template
-        single_attribute_request = single_attribute_request[:-1]
+        single_entity_operations = single_entity_operations[:-1]
```
mmorgantaylor, line 138: per other comment - remove comma from the functions so you don't have to do this

scripts/anvil_tools/batch_upsert_entities_standard.py
```diff
@@ -72,66 +96,58 @@ def create_upsert_request(tsv, array_attr_cols=None):
...
 
-        # if array columns/attributes
+        # if array columns/attributes provided
         if array_attr_cols:
             # for each array column
             for col in array_attr_cols:
-                # add column name to template for a single attribute
-                single_attribute_request += template_make_list_attr.replace("VAR_ATTRIBUTE_LIST_NAME", col)
+                # get operation json to make the array/list attribute with column name
+                single_entity_operations += create_list_attr_operation(col)
```
mmorgantaylor, line 114: as per other comment, remove comma from functions and add it here, i.e. `single_entity_operations += create_list_attr_operation(col) + ","`

scripts/anvil_tools/batch_upsert_entities_standard.py
```diff
@@ -72,66 +96,58 @@ def create_upsert_request(tsv, array_attr_cols=None):
...
                 # convert string value -> back into an array: [foo,bar] (str) --> ['foo', 'bar'] (list)
                 attr_values = convert_string_to_list(row[col])
-                # for each item in array, add row attribute value to the above created list attribute
+                # for each item in array (values pertaining to the list attribute as defined above)
                 for val in attr_values:
-                    single_attribute_request += template_add_list_member.replace("VAR_LIST_MEMBER", val).replace("VAR_ATTRIBUTE_LIST_NAME", col)
+                    # get operation json to add each array/list attribute's values to the request
+                    single_entity_operations += add_list_member_operation(col, val)
```
mmorgantaylor, line 120: as per other comments, remove comma from functions and add here

PR #62, opened 2021-07-14
Adding number of weeks from creation before a file will be mopped

meganshand:
@mmorgantaylor @samanehsan

Here is a PR to add a "time before deletion" argument to the mop script. Note that @mmorgantaylor's notebook will now need to be updated (to include a `weeks_old` argument to all the `mop` calls). 

I'm sorry for all of the whitespace changes, I think I opened it in IntelliJ and that happened automatically. I can strip them back out if that's helpful.

### DIFF
--- a/scripts/mop_workspace/mop_workspace.py
+++ b/scripts/mop_workspace/mop_workspace.py
@@ -19,12 +19,13 @@ from six import string_types #, iteritems, itervalues, u, text_type
 # from six.moves import input
 from toolz.itertoolz import partition_all
 from tqdm import tqdm
+from datetime import timedelta, datetime
 
 from firecloud import api as fapi
 
 
 def _entity_paginator(namespace, workspace, etype, page_size=500,
-                            filter_terms=None, sort_direction="asc"):
+                      filter_terms=None, sort_direction="asc"):
     """Pages through the get_entities_query endpoint to get all entities in
        the workspace without crashing.
     """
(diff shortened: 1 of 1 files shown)


mmorgantaylor (approved):
looks great!!

scripts/mop_workspace/mop_workspace.py
```diff
@@ -309,13 +313,17 @@ def update_referenced_files(referenced_files, attrs, bucket_prefix):
...
         if get_parent_directory(f) in referenced_directories:
             sibling_referenced_files.append(f)
-        
+
     # remove them from the list of unreferenced files (treat them as referenced)
     unreferenced_files = unreferenced_files - set(sibling_referenced_files)
 
     # Filter out files like .logs and rc.txt
-    def can_delete(f):
+    def can_delete(f, weeks_old_before_delete):
```
samanehsan, line 321: I think this would be easier to read if `can_delete` was defined outside of this function, but I wouldn't consider it required for this change! Just a stylistic preference :)

samanehsan (approved):
Looks great! 🕐

scripts/mop_workspace/mop_workspace.py
meganshand: I agree, but I think I'll leave it as is since that seems to be the style throughout the whole script. I'll leave it up to @mmorgantaylor for a future PR.

PR #63, opened 2021-08-17
DFE-563 Workshop Project Tool Update

Marianie-Simeon:
#Addressing
https://broadworkbench.atlassian.net/browse/DFE-563

Giving Anvil a tool for Workshop that can easily add and remove users from a workspace

# Code
Created two scripts, one to add users to the billing project and the other to remove them.


scripts/anvil_tools/README.md
```diff
@@ -12,6 +12,25 @@
...
 
+#### **add_user_to_project.py**
+##### Description
+    Add a list of users' emails to a billing project. 
+    
+    Input is a .csv file with 1 required column, other columns can be added but will be ignored by this script:
+        1. "email"
+##### Usage
+    Locally
+        `python3 /scripts/anvil_tools/add_user_to_project.py -c CSV_FILE`
```
mmorgantaylor, line 23: does the user need to add `-p BILLING_PROJECT` as well?

scripts/anvil_tools/README.md
```diff
@@ -12,6 +12,25 @@
...
+    Docker
+        `docker run --rm -it -v "$HOME"/.config:/.config -v "$HOME"/local_data_directory/:/data broadinstitute/horsefish bash -c "cd data; python3 /scripts/anvil_tools/add_user_to_project.py -c /data/INPUT.csv -p BILLING_PROJECT [-v]"`
+
+        Note: `local_data_directory` should be the path to the folder where your desired input .csv file is located.
+##### Flags
+    1. `--verbose`, `-v`: set for more detailed/information to stdout (default = False)
+    2. `--csv`, `-c`: input .csv file (required)
+    3. `--project`, `-p`: billing project name
```
mmorgantaylor, line 31: is billing project name required as well? (probably should be)

scripts/anvil_tools/README.md
```diff
@@ -119,6 +138,24 @@
...
 
+#### **remove_users_from_project.py**
+##### Description
+    Remove a list of users' emails to a billing project. 
+    
+    Input is a .csv file with 1 required column, other columns can be added but will be ignored by this script:
+        1. "email"
+##### Usage
+    Locally
+        `python3 /scripts/anvil_tools/remove_users_from_project.py -c CSV_FILE`
```
mmorgantaylor, line 149: same comment here about the billing project flag

scripts/anvil_tools/README.md
```diff
@@ -119,6 +138,24 @@
...
+    Docker
+        `docker run --rm -it -v "$HOME"/.config:/.config -v "$HOME"/local_data_directory/:/data broadinstitute/horsefish bash -c "cd data; python3 /scripts/anvil_tools/remove_users_from_project.py -c /data/INPUT.csv -p BILLING_PROJECT [-v]"`
+
+        Note: `local_data_directory` should be the path to the folder where your desired input .csv file is located.
+##### Flags
+    1. `--verbose`, `-v`: set for more detailed/information to stdout (default = False)
+    2. `--csv`, `-c`: input .csv file (required)
+    3. `--project`, `-p`: billing project name
```
mmorgantaylor, line 157: same comment here about billing project should be required

scripts/anvil_tools/add_user_to_project.py
```diff
@@ -0,0 +1,87 @@
...
+    """Adding list of users' emails to the billing project"""
+    if verbose:
+        print(email_list)
+
+    # Get access token and and add to headers for requests.
+    # -H  "accept: */*" -H  "Authorization: Bearer [token] -H "Content-Type: application/json"
+    headers = {"Authorization": "Bearer " + get_access_token(), "accept": "*/*", "Content-Type": "application/json"}
+    
+    # Stopping running code at the 6th error to avoid multiple failures
+    error_count = 0
```
mmorgantaylor, line 27: I'd also count successful emails added so you can report on how many were added

scripts/anvil_tools/add_user_to_project.py
```diff
@@ -0,0 +1,87 @@
...
+        if verbose:
+            print(f"Successfully added the email {email} to the billing project {project}.")
+    
+    # Success of Fail Messages
+    if error_count < 1:
+        print(f"Successfully added all emails to the billing project: {project}.")
+    elif error_count == len(email_list):
+        print(f"Failed to added emails to the billing project: {project}.")
+    else:
+        print(f"Successfully added some emails to the billing project: {project}.")
```
mmorgantaylor, line 61: if you add a count of emails successfully added, that would be very useful to add to this message, e.g. 
`print(f"Successfully added only {n_emails_added} emails out of {len(email_list)} to the billing project: {project}.")`

scripts/anvil_tools/add_user_to_project.py
```diff
@@ -0,0 +1,87 @@
...
+        # adding fail message
+        if status_code != 200:
+            error_count+=1
+            print(f"WARNING: Failed to add the email {email} to the billing project {project}.")
+            print("Please see full response for error:")
+            print(response.text)
+             
+            if error_count > 6:
+                print("Retry Error: Failed to Run over 6 times")
+                return "Retry Error"
```
mmorgantaylor, line 49: does this email actually get retried, or does this just continue on the loop? I think "retry error" is misleading

scripts/anvil_tools/remove_users_from_project.py
```diff
@@ -0,0 +1,89 @@
...
+
+    for email in email_list:
+        # URL Encoding the email
+        url_encode_email = urllib.parse.quote(email)
+
+        # Library/putLibraryMetadata
+        uri = f"https://rawls.dsde-prod.broadinstitute.org/api/billing/v2/{project}/members/user/{url_encode_email}"
+
+        # capture response from API and parse out status code
+        response = requests.delete(uri, headers=headers)
```
mmorgantaylor, line 39: if this code is identical to the other script except for this line, I'd just make it one script (like `modify_users_on_project.py` and have a flag for adding versus removing users to the project

mmorgantaylor (changes requested):
great work! I think these scripts should be refactored or combined since there is so much repeated code. 
+ a few smaller suggestions / questions

scripts/anvil_tools/remove_users_from_project.py
mmorgantaylor: or at least refactor all the repeated code into reusable functions

scripts/anvil_tools/add_or_remove_user_from_project.py
```diff
@@ -0,0 +1,111 @@
...
+            if verbose:
+                print(f"Successfully {action_past} the email {email} to the billing project {project}.")
+    
+    # Success of Fail Messages
+    if error_count < 1:
+        print(f"Successfully {action_past} all emails to the billing project: {project}.")
+    elif error_count == len(email_list):
+        print(f"Failed to {action} emails to the billing project: {project}.")
+    else:
+        print(f"Successfully {action_past} only {emails_added} emails out of {len(email_list)} to the billing project: {project}.")
```
mmorgantaylor, line 76: as I think about this more, I think also printing out the list of emails that failed to {action} would be useful to the user. I know they're printed out as they go, but I worry the error messages might overwhelm the user from pulling out the emails that failed.

scripts/anvil_tools/add_or_remove_user_from_project.py
```diff
@@ -0,0 +1,111 @@
...
+        # URL Encoding the email
+        url_encode_email = urllib.parse.quote(email)
+        
+        # Library/putLibraryMetadata
+        uri = f"https://rawls.dsde-prod.broadinstitute.org/api/billing/v2/{project}/members/user/{url_encode_email}"
+        
+        # capture response from API
+        if add_action:
+            response = requests.put(uri, headers=headers)
+            action="add"
```
mmorgantaylor, line 42: style nit: here and in lots of other places, please add spaces before and after `=` or `+=`

scripts/anvil_tools/add_or_remove_user_from_project.py
```diff
@@ -0,0 +1,111 @@
...
+
+    # Assigning project variable
+    project = args.project
+
+    # Assigning csv variable
+    csv = args.csv
+
+    # Assigning add_action variable
+    if args.add:
+        add_action = True
```
mmorgantaylor, line 102: if a user includes both --add and --remove, this currently defaults to add. is that your intended behavior? i might add an extra check for that; also if the user includes neither flag, they'll get an error about `add_action` not being defined; i might add a check and explicit messaging for that case as well.

scripts/anvil_tools/add_or_remove_user_from_project.py
```diff
@@ -0,0 +1,111 @@
...
+def get_access_token():
+    """Get access token."""
+
+    scopes = ["https://www.googleapis.com/auth/userinfo.profile", "https://www.googleapis.com/auth/userinfo.email"]
+    credentials = GoogleCredentials.get_application_default()
+    credentials = credentials.create_scoped(scopes)
+    return credentials.get_access_token().access_token
+
+
+def add_or_remove_user_from_project(project, email_list, add_action, verbose):
```
mmorgantaylor, line 17: great work, merging the two actions into one variable is a nice solution!

mmorgantaylor (approved):
I have a few further minor suggestions, but otherwise LGTM!

scripts/anvil_tools/add_or_remove_user_from_project.py
Marianie-Simeon: the args parser will not let them have both or neither since it's mutually exclusive, Should I still but in the check just incase?

scripts/anvil_tools/add_or_remove_user_from_project.py
Marianie-Simeon: Sure, I will add it now

scripts/anvil_tools/add_or_remove_user_from_project.py
mmorgantaylor: yes, I would put in a manual check, something like
`if args.add and args.remove: <some warning with useful instruction and exit>`
and 
`if not args.add and not args.remove: <some warning with useful instruction and exit>`

scripts/anvil_tools/add_or_remove_user_from_project.py
Marianie-Simeon: Sure, I will add it now

scripts/anvil_tools/add_or_remove_user_from_project.py
Marianie-Simeon: Ok, I will add it now. Thank you for helping me <3

mmorgantaylor: no problem! really nice job with this!

PR #64, opened 2021-09-10
[DFE-594] get a single entity table from a list of workspaces and concatenate together into an excel report

schaluva:
From a previous script - `split_and_push_data_model_tsvs.py` - a single file formatted as a Terra load table contains extra columns for workspace name and workspace project. Using this information, the single file is split and then uploaded as the appropriate table in its correct workspace.

This script does the opposite - gathers the given entity table from each of the workspaces listed and concatenates them into the format that is input into `split_and_push_data_models.py`.

A single excel file is input and a single excel file is output - the output excel file has two sheets where one is the list of workspaces that may have failed and the second is a single load tsv table format that represents the input of the script mentioned above.

### DIFF
--- a/scripts/anvil_tools/gather_and_concatenate_data_model_tsvs.py
+++ b/scripts/anvil_tools/gather_and_concatenate_data_model_tsvs.py
@@ -66,7 +66,12 @@ def gather_and_concatenate_data_model_tsvs(input_file, entity_name):
     failed_data.to_excel(writer, sheet_name="failed_workspaces", index=None)
     writer.save()
 
-    print(f"Successfully completed gather and concatenate. All results can be found in {output_filename}.")
+    # if any failures, print warning message.
+    if len(failed_workspaces) > 0:
+        print(f"Warning: Completed gather and concatenate with the exception of some workspace/s. Please examine details in {output_filename}.")
+        return
+    # else print success message
+    print(f"Successfully completed gather and concatenate for all workspaces. Results can be found in {output_filename}.")
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


scripts/anvil_tools/gather_and_concatenate_data_model_tsvs.py
```diff
@@ -0,0 +1,81 @@
...
+    """Get data table tsv files from list of workspaces and concatenate results into a single excel report."""
+
+    # read full excel sheet into dataframe - all rows of workspace project and workspace names
+    workspace_info = pd.read_excel(input_file, sheet_name="Sheet1", index_col=None)
+
+    # instantiate empty list to hold all entity information from all workspaces
+    all_workspace_entities = []
+    failed_workspaces = []
+    # for each workspace_name, workspace_project pair
+    for index, workspace in workspace_info.iterrows():
```
mmorgantaylor, line 22: does this break if there are not exactly 2 columns in the input file? is there any check to make sure the input file is formatted correctly with the right order of columns?

scripts/anvil_tools/gather_and_concatenate_data_model_tsvs.py
```diff
@@ -0,0 +1,81 @@
...
+
+    # # write final dataframes to excel file - separate sheets for success and failed data
+    output_filename = input_file.split("/")[-1].split(".")[0] + "_final.xlsx"
+    writer = pd.ExcelWriter(output_filename, engine="openpyxl")
+
+    succeeded_data.to_excel(writer, sheet_name="concatenated_entity_table", index=None)
+    failed_data.to_excel(writer, sheet_name="failed_workspaces", index=None)
+    writer.save()
+
+    print(f"Successfully completed gather and concatenate. All results can be found in {output_filename}.")
```
mmorgantaylor, line 69: add optional messaging if there were errors?

mmorgantaylor (approved):
looks great! a couple minor questions/comments

scripts/anvil_tools/gather_and_concatenate_data_model_tsvs.py
schaluva: so order of the columns shouldn't matter since im getting columns with their exact names. I can test it but it also should be able to have additional columns but at minimum, has the two that the script is looking to parse workspace information from.

PR #68, opened 2021-10-01
TDR ingest data and metadata scripts

schaluva:


scripts/emerge/one_step_data_ingest.py
```diff
@@ -0,0 +1,57 @@
...
+
+    all_row_jsons = []
+    for key in row_json.keys():
+        value = str(row_json[key])  # convert to string to be able to check if gs:// path
+
+        # if value exists and has 'gs://'
+        if value is not None and "gs://" in value:
+            relative_tdr_path = value.replace("gs://","/")  # create TDR relative path
+
+            # # TODO: add in description = id_col + col_name
```
kcibul, line 17: I was curious where description even shows up?

scripts/emerge/one_step_data_ingest.py
```diff
@@ -0,0 +1,57 @@
...
+            row_json[key] = {"sourcePath":value,
+                             "targetPath":relative_tdr_path,
+                             "mimeType":"text/plain"
```
kcibul, line 20: I wasn't even setting this (I'm not sure why they ask or what the effect is)… and why TDR isn't reading mimeType off the imported file itself from GCS

scripts/emerge/one_step_data_ingest.py
```diff
@@ -0,0 +1,57 @@
+import argparse
+import json
+import pandas as pd
+
+
+def create_recoded_json(row_json):
+    """Update dictionary with TDR's dataset relative paths for keys with gs:// paths."""
+
+    all_row_jsons = []
```
kcibul, line 9: unused?

scripts/emerge/one_step_data_ingest.py
schaluva: apparently its in the file metadata when you ask to get that back

scripts/emerge/one_step_data_ingest.py
schaluva: im not sure either but last time i left it out i was told it was required so i just added it in

scripts/emerge/deprecated/create_file_ingest_json.py
```diff
@@ -0,0 +1,26 @@
...
+def create_file_ingest_json(input_file):
+    """Create newline delimited json file for TDR bulk file ingest API."""
+
+    ingest_files = pd.read_excel(input_file, sheet_name="Sheet1", index_col=None)
+
+    output_basename = input_file.split(".xlsx")[0]
+    output_json = open(f"{output_basename}_newline_delimited.json", "w")
+
+    for row in ingest_files.iterrows():
+        row[1].to_json(output_json)
```
mmorgantaylor, line 14: why are you grabbing the second item in `row` ?

schaluva: `iterrows()` returns an object (Series, I think) with the index of each row and the data in each row so the second item is the data for each row. Probably could just handle this without needing this but I stopped developing this script early on. I put it into a deprecated folder just to not have to delete it in case I wanted to reference it for code snippets.

PR #102, opened 2022-07-13
update with emerge_id and poly_id - do not delivery crsp

schaluva:


mmorgantaylor (approved):
my only suggestion would be that somewhere you define a variable called like `field_name_for_id` that you set for now to "anvil_poly_sample_id" and then use that everywhere you have that string, rather than repeating the string lots of places. it will make future refactors like this easier, and avoids issues from accidentally forgetting to update some instances.

PR #105, opened 2022-08-27
Update path for resources directory

ncalvanese1:
Updating the path for the resources directory (it is a sub-directory of the ingest_pipeline directory).


schaluva (approved):
was that in there incorrectly from the start or did the directory structure change? just curious is all!

PR #106, opened 2022-08-29
update script with errors and more stdout

schaluva:


scripts/emerge/ingest_to_tdr.py
```diff
@@ -103,29 +98,37 @@ def call_ingest_dataset(control_file_path, target_table_name, dataset_id):
...
-        return
+        raise ValueError(error_message)
 
     print("File ingest to TDR dataset completed successfully.")
 
 
-def write_load_json_to_bucket(bucket, recoded_rows_json, timestamp):
-    """Write the list of recoded rows (list of dictionaries) to a file and copy to workspace bucket."""
+def write_load_json_to_bucket(bucket, filename):
+    """Cope newline delimited json file to workspace bucket."""
```
mmorgantaylor, line 107: typo: cope -> copy

scripts/emerge/ingest_to_tdr.py
```diff
@@ -180,26 +183,24 @@ def create_recoded_json(row_json):
...
     tsv_df = pd.read_csv(input_tsv, sep="\t")
+    all_recoded_row_dicts = []
 
-    basename = input_tsv.split(".tsv")[0]
-    output_filename = f"{basename}_recoded_newline_delimited.json"
-
-    all_rows = []
     for index, row in tsv_df.iterrows():
-        # create timestamp
+        # create output filename
```
mmorgantaylor, line 193: is this comment accurate? i don't see a filename being created

scripts/emerge/resolve_drs_uris.py
```diff
@@ -0,0 +1,73 @@
...
+
+def resolve_drs_to_gs_paths(infile, outfile_name):
+    """For each line in file, resolve drs uri to gs path."""
+
+    with open(infile, 'r') as input:
+        drs_paths_to_resolve = input.readlines()
+    
+    gs_paths = []
+    for drs in drs_paths_to_resolve:
+        drs_uuid = drs.split("/")[3].strip("\n")
```
mmorgantaylor, line 58: minor documentation request: can you comment an example of what `drs` is to begin with so we understand better why you're pulling out the 4th value of the split?

scripts/emerge/resolve_drs_uris.py
```diff
@@ -0,0 +1,73 @@
...
+        for item in input_list:
+            outfile.write(f"{item}\n")
+
+
+def resolve_drs_uri(drs_uuid):
+    """Resolve a drs uri to return the corresponding gs path."""
+
+    print(f"resolving {drs_uuid}")
+    uri = f"https://data.terra.bio/ga4gh/drs/v1/objects/{drs_uuid}"
+    print(uri)
```
mmorgantaylor, line 35: just checking: do you want to keep all these print statements in?

scripts/emerge/resolve_drs_uris.py
```diff
@@ -0,0 +1,73 @@
...
+
+    headers = {"Authorization": "Bearer " + get_access_token(), "accept": "*/*"}
+
+    response = requests.get(uri, headers=headers)
+    status_code = response.status_code
+
+    if status_code != 200:
+        return response.text
+
+    gs_path = response.json()["access_methods"][0]["access_url"]["url"]
```
mmorgantaylor, line 45: does `access_methods` include a list of items? how do we know we only want the first one? what if the first one isn't a google path?

mmorgantaylor (approved):
Mostly minor questions/suggestions, one possibly more important question about drs resolution and whether you can count on there being a google path to resolve to

scripts/emerge/ingest_to_tdr.py
schaluva: ah yeah it used to exist there but i moved around a few things and left that behind

scripts/emerge/resolve_drs_uris.py
```diff
@@ -0,0 +1,80 @@
...
+    avail_access_methods = response.json()["access_methods"]
+    # if there is an access method with "type" == "gs", get the gs url (other type options = https)
+    gs_url = [access_type["access_url"]["url"] for access_type in avail_access_methods if access_type["type"] == "gs"] # ["gs://"] or []
+
+    # if a gs url is not available, return "NA"
+    if not gs_url:
+        return "NA"
+    
+    # if gs url is available, return gs path
+    return gs_url[0]
```
mmorgantaylor, line 53: i suggest checking that the length of `gs_url` is 1, and if it's not, at least printing a warning that multiple paths were found. i don't know if this is possible, but it would be good to flag it in case it does happen.

scripts/emerge/resolve_drs_uris.py
```diff
@@ -0,0 +1,80 @@
...
+        return response.text
+
+    # get a list of the available access methods for a drs uri
+    avail_access_methods = response.json()["access_methods"]
+    # if there is an access method with "type" == "gs", get the gs url (other type options = https)
+    gs_url = [access_type["access_url"]["url"] for access_type in avail_access_methods if access_type["type"] == "gs"] # ["gs://"] or []
+
+    # if a gs url is not available, return "NA"
+    if not gs_url:
+        return "NA"
```
mmorgantaylor, line 50: is the string "NA" for sure what you want to be written to the file? will the output file only be human-read? if it's ever going to be used as input to another script, or uploaded to Terra, I might suggest setting this to `None` instead of the "NA" string.

mmorgantaylor (approved):
nice! a couple small comments for you to take or leave!

scripts/emerge/resolve_drs_uris.py
mmorgantaylor: yeah makes sense! i think either commenting that we are confident about the assumption that there will only ever be one (once someone from jade confirms), or adding an assertion or check (possibly in either case, but it's really up to you), would be great.

PR #112, opened 2022-09-28
[DFE-784] Script to set up a new WFL module for Twist-TCAP datasets

mmorgantaylor:
see README for documentation.

I've tested this script functionally (ran it, then pushed data to the dataset and saw that WFL picked up the data and kicked off the appropriate workflow).


scripts/wfl/twisttcap/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+COPY requirements.txt .
+RUN apt-get install python3-pip
+RUN pip3 install -r requirements.txt
+
+COPY setup_new_wfl_module.py /scripts/setup_new_wfl_module.py
+
+ENV PYTHONPATH "/scripts:${PYTHONPATH}"
+
+CMD ["/bin/bash"]
```
kyuksel, line 14: 1. Not necessarily applicable to this Dockerfile but a general [FYI on multi-stage builds](https://towardsdatascience.com/using-multi-stage-builds-to-make-your-docker-image-almost-10x-smaller-239068cb6fb0) in case you weren't aware of that feature
2. Another good (security) practice is to avoid running as `root` whenever possible ([reference](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user)).

scripts/wfl/twisttcap/setup_new_wfl_module.py
```diff
@@ -0,0 +1,220 @@
...
+import requests
+import google.auth as googleauth
+from firecloud import api as fapi
+from time import sleep
+from pprint import pprint
+
+# DEVELOPER: update this field anytime you make a new docker image
+docker_version = "1.0"
+
+
```
kyuksel, line 12: Another FYI that's probably not very applicable to this file on using [Python type hints](https://www.pythontutorial.net/python-basics/python-type-hints), which becomes more and more valuable as Python modules get more complicated and interconnected although the cut-off would probably vary considerably from software eng to software eng 😊

kyuksel (commented):
Some general comments and references. Sorry if you already knew about them.

scripts/wfl/twisttcap/setup_new_wfl_module.py
mmorgantaylor: ah I'd love to start doing this. would you recommend using `mypy` like that article suggests?

ahaessly (approved):
👍

scripts/wfl/twisttcap/Dockerfile
ahaessly: good point kaan, I'm going to post a reference to this article in the slack channel

mmorgantaylor: @kyuksel can you give an example of a Dockerfile not running as `root`? thanks! (and thanks Andrea for adding the link in Slack - figured that one might involve more of a discussion/troubleshooting so didn't want to hash it out in a long github thread)

scripts/wfl/twisttcap/setup_new_wfl_module.py
kyuksel: `mypy` is what I used but it was over 3 years ago so not sure if there may be a better tool now but probably no need to make perfect the enemy of good even if there is 😊

scripts/wfl/twisttcap/Dockerfile
kyuksel: @mmorgantaylor Sure. [This link](https://medium.com/@mccode/processes-in-containers-should-not-run-as-root-2feae3f0df3b) has a good illustrative example (and also explains reasonings in a more friendly way than the first reference I posted above) 😊 Going with that example: Instead of

```Dockerfile
FROM debian:stretch
CMD ["cat", "/tmp/secrets.txt"]
```

we would have

```Dockerfile
FROM debian:stretch
RUN groupadd -g 999 appuser && \
    useradd -r -u 999 -g appuser appuser
USER appuser
CMD ["cat", "/tmp/secrets.txt"]
```

and more generally

```Dockerfile
FROM <base image>
RUN groupadd -g 999 appuser && \
    useradd -r -u 999 -g appuser appuser
USER appuser
... <rest of Dockerfile> ...
```

Sometimes there are operations to perform that _requires_ `root`. In those cases, we run as `root` only when required and the switch back to `non-root`. For [example](https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-base/Dockerfile), [here](https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-base/Dockerfile#L5) we become `root`, do a bunch of `root` things, and finally [here](https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-base/Dockerfile#L152) we become `non-root` and leave it that way so when someone runs a container using an image created from that `Dockerfile`, they run as `non-root`.

Does ^^^ clarify?

scripts/wfl/twisttcap/find_running_wfl_workloads.py
```diff
@@ -0,0 +1,93 @@
+import argparse
+import json
+import requests
+import google.auth as googleauth
+from firecloud import api as fapi
+from time import sleep
+from pprint import pprint
+
+
+def get_access_token():
```
kilonzi, line 10: Suggestion to have these two functions get_access_token and get_headers in their own _file.py_ and not repeated in the two files.

scripts/wfl/twisttcap/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM gcr.io/google.com/cloudsdktool/google-cloud-cli:404.0.0-slim
```
kilonzi, line 1: For own knowledge: Is it possible to start from a different docker image say 
> FROM python:3.8-slim-buster
and install the `cloudsdk` through PIP or APT-GET? The potential would be a smaller image output

scripts/wfl/twisttcap/find_running_wfl_workloads.py
mmorgantaylor: oh great call, definitely want to do this. thanks!

scripts/wfl/twisttcap/Dockerfile
mmorgantaylor: yes, very helpful! if I add that, then I don't have permission to run some of the other processes in my Dockerfile. hopefully we can work on this together sometime soon.

kyuksel: Would it work if you put

```Dockerfile
RUN groupadd -g 999 appuser && \
    useradd -r -u 999 -g appuser appuser
USER appuser
```

right before 

```Dockerfile
CMD ["/bin/bash"]
```

?

Working together sounds good too.

mmorgantaylor: this works to build it, but when I go to push to dockerhub, I get ```denied: requested access to the resource is denied```

kyuksel: To make sure I understand, you're pushing the image you built from this Dockerfile, to DockerHub, on your laptop, correct? If so, I'm not sure how that'd be related to how the image is built. I wonder [docker login](https://forums.docker.com/t/docker-push-error-requested-access-to-the-resource-is-denied/64468/2)ing would help. 

P.S. I haven't used DockerHub in a while because I've heard several times from folks like Lukas Karlsson that the Broad wanted us to move away from it for various reasons. I believe [Artifact Registry](https://cloud.google.com/artifact-registry) is the recommended one.

PR #113, opened 2022-10-05
Add loadtag ingest

schaluva:


scripts/emerge/ingest_to_tdr.py
```diff
@@ -21,84 +21,116 @@ def get_access_token():
...
+    ingest_dataset_request = create_ingest_dataset_request(control_file_path, target_table_name, load_tag) # create request for ingestDataset
+    print(f"ingestDataset request body: \n {ingest_dataset_request} \n")
+
+    ingest_response = ingest_dataset(dataset_id, ingest_dataset_request) # call ingestDataset
+    print(f"ingestDataset response body: \n {ingest_response} \n")
+
+    ingest_job_id = ingest_response["id"] # check for presence of id in ingest_dataset()
+    ingest_status_code = ingest_response["status_code"]
+
+    if ingest_status_code != 202: # job submitted but fails
```
mmorgantaylor, line 105: didn't you already check for this inside `ingest_dataset` ? (line 68)

i think you can delete these 3 lines

scripts/emerge/ingest_to_tdr.py
```diff
@@ -21,84 +21,116 @@ def get_access_token():
...
+
+    while job_status_code == 202:           # while job is running
+        print(f"{ingest_job_id} --> running")
+        time.sleep(10) # wait 10 seconds before getting job status
+        job_status_code, job_status_response = get_job_status(ingest_job_id) # get updated status info
+
+    # job completes (‘failed’ or ‘succeeded’) with 200 status_code
+    # consider any combination other than 200 + succeeded as failure
+    if job_status_code != 200 or job_status_response["job_status"] != "succeeded":
+        print(f"{ingest_job_id} --> failed")
```
mmorgantaylor, line 120: hmm if job_status_code is not 200, if it's e.g. 500 (something wrong with the service transiently) that doesn't necessarily mean the job failed, right? this might be enough of an edge case not to worry about but i think the logic here isn't quite water tight

scripts/emerge/ingest_to_tdr.py
```diff
@@ -21,84 +21,116 @@ def get_access_token():
...
+def create_ingest_dataset_request(control_file_path, target_table_name, load_tag=None):
+    """Create the ingestDataset request body."""
+
+    load_dict = {"format": "json",
+                 "path": control_file_path,
+                 "table": target_table_name,
+                 "resolve_existing_files": "true",
+                 "updateStrategy": "replace"
+                }
+    # if user provides a load_tag, add it to request body
```
mmorgantaylor, line 84: what does including a load tag do?

mmorgantaylor (approved):
looks good, just some questions and possibly some cleanup

scripts/emerge/ingest_to_tdr.py
schaluva: yeah i wanted to handle it "simple" in that if its not a true true success - aka the 200 and a "success" then its considered as a fail so as not to run the risk of failed ingests into TDR. Aka re-run it if it fails for any reason other than fully successful. sort of naive but i think operating in a "better safe than sorry" method here for emerge since its a larger scale number of ingests.

scripts/emerge/ingest_to_tdr.py
mmorgantaylor: cool, thanks! i think our future selves would benefit from a comment somewhere to explain this briefly, maybe in the argparse section, like "allows ingest of duplicate files in separate ingest calls when using the same load tag"

scripts/emerge/ingest_to_tdr.py
mmorgantaylor: that makes sense - but in that case then i'm not sure that what you wrote in row 126 is guaranteed to be true

if you want it to be true then i'd change row 119 to
`if not (job_status_code == 200 and job_status_response["job_status"] == "succeeded"):`

scripts/emerge/ingest_to_tdr.py
schaluva: ah good catch.

scripts/emerge/ingest_to_tdr.py
schaluva: smart using the `not`

i was playing around with
```
    if job_status_code == 200 and job_status_response["job_status"] == "succeeded":
        # job completes successfully
        # success_code is 200 and "job_status" =is"succeeded"
        print(f"{ingest_job_id} --> succeeded")

print(f"{ingest_job_id} --> failed")
# if failed, get the resulting error message
job_result_code, job_result_response = get_job_result(ingest_job_id)
raise ValueError(job_result_response)   
``` 
but i didnt want the flow of conditionals to flip around from what it was above.

thanks for the suggestions it sorts out what i was trying to get to

PR #114, opened 2022-10-12
[SU-202] Add new Terra service banner scripts to add/remove multiple banners

MatthewBemis:


scripts/service_alerts/add_service_alert.py
```diff
@@ -0,0 +1,47 @@
+import argparse
+from google.cloud import storage as gcs
+import json
+from utils import *
```
mmorgantaylor, line 4: minor: generally prefer to explicitly name all imports rather than use `import *`

mmorgantaylor (approved):
looks great to me! one minor request about imports.

scripts/service_alerts/add_service_alert.py
MatthewBemis: Updated in 0e41967a1dcf755f16042f8e5c1ec94e40125c51

nawatts (approved):

PR #115, opened 2022-10-19
add project for requester pays to copy anvil data to delivery workspace

schaluva:


scripts/emerge/stage_prs_anvil_deliverable_data.py
```diff
@@ -108,7 +108,13 @@ def rename_and_rehome_data_files(prs_entities_dicts, dest_bucket, snapshot_id):
             src_blob = "/".join(prs_entity[file_type].split("/")[3:]) # TDR source filepath
 
             print(f"Initiate copy of gs://{src_bucket}/{src_blob} to {dest_filepath}")
-            copy_object(src_bucket, src_blob, dest_bucket, dest_blob)
+
+            # if project provided for requester pays
+            if project_id:
```
mmorgantaylor, line 113: I don't think you need this if statement. if project_id isn't input to this function, it defaults to `None`, so you should be able to just call `copy_object(src_bucket, src_blob, dest_bucket, dest_blob, project_id)` and it will work whether project_id is defined or not

mmorgantaylor (approved):
looks great!

PR #117, opened 2022-10-25
[DFE-790] update mop to support submissions path

mmorgantaylor:
Our more scaleable version of mop (in mmt_add_mop branch of horsefish repo) only looks in the old submissions directories, i.e. `gs://bucket/submissionid/...`, but Terra now stores submissions in a new format: `gs://bucket/submissions/submissionid/...` so we're now missing files from recent submissions and failing to clean them up.

This PR updates the code to look in both places

### DIFF
--- a/scripts/mop_workspace/mop_workspace.py
+++ b/scripts/mop_workspace/mop_workspace.py
@@ -127,10 +127,17 @@ def list_bucket_files(project, bucket_name, referenced_files, verbose):
             return None
 
         full_file_path = "gs://" + bucket_name + "/" + blob_name
-        # Splits the bucket file: "gs://bucket_Id/submission_id/file_path", by the '/' symbol
-        # and stores values in a 5 length array: ['gs:', '' , 'bucket_Id', submission_id, file_path]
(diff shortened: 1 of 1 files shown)


meganshand (approved):
👍 One question, in the case where there is a user generated directory, or a Terra generated "notebooks" directory, the submission_id will either be a non-submission value, or empty. Do those cases get filtered out somewhere?

mmorgantaylor:
@meganshand good question, yes, this script only considers submissions files - this is why it was failing to find the new ones in the first place. to generate the list of files to delete, we start with all files that we've identified as being related to a submission (that identification is what changes in this PR):
https://github.com/broadinstitute/horsefish/pull/117/files#diff-42dc123584d9d2dbda0da5fe006f31eedd48b13e696c88c67544444e8fbcbbc3R307

and then remove all files referenced by the data tables (we do not want to delete those files)

a file generated by a user or in a notebooks directory won't be associated with a submission to begin with, so it won't be a candidate for deletion.

meganshand:
I tested this branch out on a newer workspace and everything worked great. Thanks!

PR #123, opened 2023-01-09
ingest json files from bucket to respective tdr dataset tables

schaluva:
Problem:
RARE-X datasets have ~64 tables that are ingested with data from new-line delimited json files. The json files are generated by a third-party company and put into a GCP bucket for RARE-X team to ingest. Currently, someone on the RARE-X team manually ingests each table separately. 

Solution:
Script that will take GCS storage path pointing to .json files and ingest each json to its respective table in the TDR dataset. the .json files are named [table_name].json.

### DIFF
--- a/scripts/rare-x/ingest_rarex_dataset.py
+++ b/scripts/rare-x/ingest_rarex_dataset.py
@@ -176,7 +176,6 @@ if __name__ == "__main__" :
     args = parser.parse_args()
 
     # parse bucket_name and subdirectories
-    # TODO: handle if there is no subdir
     if args.bucket_path.startswith("gs://"):
         bucket_name = args.bucket_path.split("/")[2] # bucket_name
         subdir = "/".join(args.bucket_path.split("/")[3:]).strip("/") # subdirectory path in bucket
--- a/scripts/rare-x/ingest_rarex_dataset.py
+++ b/scripts/rare-x/ingest_rarex_dataset.py
@@ -1,14 +1,10 @@
 # imports and environment variables
 import argparse
 import json
-import pandas as pd
-import pytz
 import requests
 import time
 
-from datetime import datetime
 from google.cloud import storage
-from google.cloud import storage as gcs
 from oauth2client.client import GoogleCredentials
 
 
@@ -23,7 +19,7 @@ def get_access_token():
 
 
 def get_job_result(job_id):
-    """retrieveJobResult"""
+    """Call retrieveJobResult."""
 
     # first check job status - retrieveJob
     uri = f"https://data.terra.bio/api/repository/v1/jobs/{job_id}/result"
@@ -39,7 +35,7 @@ def get_job_result(job_id):
 
 
 def get_job_status(job_id):
-    """retrieveJobStatus"""
+    """Call retrieveJobStatus."""
 
     # first check job status - retrieveJob
     uri = f"https://data.terra.bio/api/repository/v1/jobs/{job_id}"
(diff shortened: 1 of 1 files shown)


scripts/rare-x/ingest_rarex_dataset.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def create_ingest_dataset_request(ingest_filepath, target_table_name):
+    """Create the ingestDataset request body."""
+
+    load_dict = {"format": "json",
+                 "path": ingest_filepath,
+                 "table": target_table_name,
+                 "resolve_existing_files": "true",
+                 "updateStrategy": "replace"
```
mmorgantaylor, line 111: do you always want to use the replace strategy? could be nice to have this be an input to the function to allow for "merge" - otherwise I'd suggest changing the name of the function to explicitly indicate that you're calling the "replace" strategy

scripts/rare-x/ingest_rarex_dataset.py
```diff
@@ -0,0 +1,187 @@
...
+                }
+
+    load_json = json.dumps(load_dict) # dict -> json
+
+    return load_json
+
+
+def run_ingest(dataset_id, data_filepaths):
```
mmorgantaylor, line 119: could you clarify what `data_filepaths` and "json path" are? are they the same thing? i think it's the path to the ingest json for a given table - in any case, making a slightly more descriptive (and consistent) variable name would help here

scripts/rare-x/ingest_rarex_dataset.py
```diff
@@ -0,0 +1,187 @@
...
+        ingest_response = ingest_dataset(dataset_id, ingest_request) # call ingestDataset
+        print(f"Submit ingest response body: \n {ingest_response} \n")
+
+        ingest_complete_response = monitor_ingest(ingest_response)
+        print(f"Finished ingest response body: \n {ingest_complete_response} \n")
+
+        print(f"Finished ingest for table {table_name}. \n\n")
+
+
+def get_json_paths(bucket_name, subdir):
```
mmorgantaylor, line 142: similar to comment above - what does "json paths" mean here?

mmorgantaylor (approved):
nice! a few comments to take or leave for code readability

scripts/rare-x/ingest_rarex_dataset.py
schaluva: yeah always replace for them

mmorgantaylor: cool, honestly I'd probably still make update_strategy an input but set the default so you don't have to change anything else, like: 

`def create_ingest_dataset_request(ingest_filepath, target_table_name, update_strategy="replace"):`

and then on this line `"updateStrategy": update_strategy`

scripts/rare-x/ingest_rarex_dataset.py
```diff
@@ -0,0 +1,189 @@
...
+        print(f"Submitted ingest response body: \n {ingest_response} \n")
+
+        ingest_complete_response = monitor_ingest(ingest_response)
+        print(f"Finished ingest response body: \n {ingest_complete_response} \n")
+
+        print(f"Finished ingest for table {table_name}. \n\n")
+
+
+def get_ingest_data_json_filepaths(bucket_name, subdir):
+    """Get list of gs:// paths for each new-line delimited json containing data to ingest into TDR dataset table."""
```
mmorgantaylor, line 143: I think I'd specifically state the logic here as it relates to the input - you're looking for all json files in the given bucket_name/subdir

mmorgantaylor (approved):
love the new variable name

PR #125, opened 2023-01-11
Update Anv ws set-up to add anvil_devs to AD

ncalvanese1:


ncalvanese1:
Updated the code to support two changes: 

1. Adding AnVIL_Devs@firecloud.org as a member on any created Auth Domains, to ensure AnVIL developers can access the dataset. 
2. Allowing the creation of workspaces for open-access datasets (where an Auth Domain should not be included). This is done by simply not naming an Auth Domain in the input tsv.

schaluva (approved):

PR #127, opened 2023-02-01
[DFE-805] Allow processing of multiple pages in Terra tables script for data import

kilonzi:
Ticket: https://broadworkbench.atlassian.net/browse/DFE-805?filter=10585

**Background**: 
Previously, a CFoS script was written to allow consortium members and Broad counterparts to upload data files to Terra data tables from a CSV. The CFoS script required all table information be located in the same sheet in the excel spreadsheet, and broke out the information by columns. It then read a JSON file which defined the Schema (which tables to create and what columns belong in what tables) to parse the excel sheet and extract the information into the corresponding tables. Once this was done, validation was run on the data ensuring that it matched the expected schema and basic data requirements.

The validation and the schema documents contained conflicting information, and there was hard-coding of certain columns which made the script less flexible to changes in the future. Additionally, the JSON file was a user input, even though no users outside the Broad project team should be able to change the requirements/schema definitions in the file. 

**Changes**:
We were asked to update the script to allow each table in the schema to be represented in a different excel sheet, rather than the same sheet. 

In addition, we updated the schema file to also include column definitions, which can be used to add/remove columns and define what validation should be applied to each column. The new column definitions are used to more flexibly parse columns from the excel sheets by reading the column into the data frame if it exists in the column definitions. This separates validation from data ingest, as we validate the columns and the table schemas later during the validation step. 

Validation has been updated to dynamically build a PandasSchema object based on the JSON file's column definitions. This is then applied to the data frame, and errors are handled the same way they were previously. 

In addition, we noticed that current data contains file_IDs, where it didn't before. The code used to construct the file_IDs, but that code was temporary and somewhat brittle. Since file IDs are included in the data directly, we have removed this code and are requiring the input data contain file_IDs from the beginning. 

- Changed schema validation by adding columns
- Changed program to read multiple sheets in Excel
- Schema validation reads from Schema config json

After validation, the data frames are saved to individual TSV files, one per table. Then, a Terra API is called to upload each of the TSV files to the appropriate workspace.


scripts/cfos/dataset_tables_schema.json
```diff
@@ -1,13 +1,202 @@
...
+                                "Transcriptomic", "SpatialTranscriptomics", "Trascriptomic_Targeted", "Trascriptomic_NonTargeted"]
+    },
+
+    "has_disease" : {
+      "field_name" : "has_disease",
+      "field_type" : "category",
+      "allowed_values": ["y", "n"]
+    },
+
+    "not_here" : {
```
kilonzi, line 147: @KutnerBroadie is this supposed to be a column?

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -87,29 +88,34 @@ def generate_load_table_files(dataset_tables_dict):
...
-        # add KVP {table name : table_df} entry to all tables dictionary
-        dataset_tables_dict[table] = table_df
-
-    print(f"Success: Dataset has been partitioned into {len(dataset_table_names)} tables: {dataset_table_names}")
-    return dataset_tables_dict
-
-
-def validate_dataset(dataset):
+def validate_dataset(dataset, schema_dict, field_dict):
+    for table in list(schema_dict.keys()):
```
kilonzi, line 92: Should this be validating the columns vis a vis entire dataset only?

kilonzi (commented):
pure comments

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: What do you mean by vis a vis entire dataset?

kilonzi: Does this function validate columns or columns and the data too?

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -143,57 +108,98 @@ def load_excel_input(excel, expected_dataset_cols):
...
+    parser.add_argument(
+        '-w', 
+        '--workspace', 
+        required=False, 
+        type=str, 
+        help='Terra workspace name.'
+    )
+    parser.add_argument(
+        '-v', 
+        '--validate',
```
mmorgantaylor, line 172: minor: I'd call this `validate_only` since its action is exclusionary and you don't need this flag if you want to validate and create TSVs

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -143,57 +108,98 @@ def load_excel_input(excel, expected_dataset_cols):
...
+        required=False, 
+        action='store_true', 
+        help='Set parameter to upload data table files to workspace. default = no upload'
+    )
 
     args = parser.parse_args()
+    schema_json = 'dataset_tables_schema.json'
+
+    # TODO: Validate that schema exists; error and quit if not. Store validation code in new requirements or file_paths validation file.
+    # ask John - why write separate validation? Why not just use a Try block?
```
mmorgantaylor, line 189: philosophical: also, why not just allow the error in the first place? if the schema doesn't exist, what happens? is whatever native error happens sufficient? what's the benefit of adding extensive checks here?

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -143,57 +108,98 @@ def load_excel_input(excel, expected_dataset_cols):
...
     args = parser.parse_args()
+    schema_json = 'dataset_tables_schema.json'
+
+    # TODO: Validate that schema exists; error and quit if not. Store validation code in new requirements or file_paths validation file.
+    # ask John - why write separate validation? Why not just use a Try block? 
+    # https://therenegadecoder.com/code/how-to-check-if-a-file-exists-in-python/#check-if-a-file-exists-with-a-try-block indicates try block is more robust.
 
     # parse schema using dataset name to get list of expected columns
-    expected_dataset_cols, dataset_table_names, schema_dict = get_expected_columns_list(args.schema, args.dataset_name)
+    schema_dict, column_dict = parse_config_file(schema_json=schema_json)
```
mmorgantaylor, line 193: minor: you don't need to specify the input variable name - `parse_config_file(schema_json)` is sufficient

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -143,57 +108,98 @@ def load_excel_input(excel, expected_dataset_cols):
...
-    
+
     print("Success: Excel file has been loaded into a dataframe.")
-    return dataset_metadata_df
+    return raw_dataset_df #dataset_metadata_df
 
 
-def get_expected_columns_list(schema_json, dataset_name):
+def parse_config_file(schema_json):
     """Get list of expected columns in excel based on selected dataset schema."""
```
mmorgantaylor, line 117: I think this function description needs to be updated (i.e. it no longer returns a list)

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -143,57 +108,98 @@ def load_excel_input(excel, expected_dataset_cols):
...
+        '-d', 
+        '--dataset_name', 
+        required=False,
+        type=str, 
+        help='Dataset type to apply validations and data table structure. ex. proteomics, atac-seq, transcriptomics, singe-cell'
+    )
+    parser.add_argument(
+        '-x', 
+        '--excel', 
+        required=False,
```
mmorgantaylor, line 152: based on its use in line 195, shouldn't this field be required?

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -87,54 +58,48 @@ def generate_load_table_files(dataset_tables_dict):
...
-
-
-def load_excel_input(excel, expected_dataset_cols):
+def load_excel_input(excel, allowed_dataset_cols):
     """Read excel file input into a dataframe and validate for required columns."""
 
     try:
-        dataset_metadata_df = pd.read_excel(excel, sheet_name="Sheet1", skiprows=2, usecols=expected_dataset_cols, index_col=None)
-
+        raw_dataset_df = pd.read_excel(excel, sheet_name=None, skiprows=2, index_col=None, usecols=lambda x: x in allowed_dataset_cols)
```
mmorgantaylor, line 100: why skiprows=2? is there a description somewhere of the expected formatting of the excel input?

mmorgantaylor: also it looks like usecols can be a list, so I don't think you need a lambda here - you can just have `usecols=allowed_dataset_cols`. https://pandas.pydata.org/docs/reference/api/pandas.read_excel.html

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -143,57 +108,98 @@ def load_excel_input(excel, expected_dataset_cols):
         else:
             print(e)
         exit()
-    
+
     print("Success: Excel file has been loaded into a dataframe.")
-    return dataset_metadata_df
+    return raw_dataset_df #dataset_metadata_df
```
mmorgantaylor, line 113: cleanup: remove commented old variable

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -143,57 +108,98 @@ def load_excel_input(excel, expected_dataset_cols):
...
     # parse schema using dataset name to get list of expected columns
-    expected_dataset_cols, dataset_table_names, schema_dict = get_expected_columns_list(args.schema, args.dataset_name)
+    schema_dict, column_dict = parse_config_file(schema_json=schema_json)
     # load excel to dataframe validating to check if all expected columns present
-    dataset_metadata = load_excel_input(args.excel, expected_dataset_cols)
+    dataset_metadata = load_excel_input(args.excel, list(column_dict.keys()))
 
     # if validation flag, only validate
     if args.validate:
         validated_dataset = validate_dataset(dataset_metadata)
```
mmorgantaylor, line 199: you need to update this invocation of the function to include the additional new input args (schema_dict, tsv_filenames_list)

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -143,57 +108,98 @@ def load_excel_input(excel, expected_dataset_cols):
...
     else:
-        validated_dataset = validate_dataset(dataset_metadata)
-        dataset_tables_dict = create_dataset_tables_dictionary(validated_dataset, args.dataset_name, dataset_table_names, schema_dict)
-        tsv_filenames_list = generate_load_table_files(dataset_tables_dict)
+        validated_dataset = validate_dataset(dataset=dataset_metadata, schema_dict=schema_dict[args.dataset_name], field_dict=column_dict)
```
mmorgantaylor, line 201: minor: again here it's not necessary to name each input arg, especially if the variables are reasonably named. i.e. you could rename column_dict to field_dict.

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -143,57 +108,98 @@ def load_excel_input(excel, expected_dataset_cols):
...
+        # if the value error is about missing columns in input file
+        exit(e)
+    
 
-    # get all cols (attributes) from all tables for dataset
-    # returns nested list of lists with column names
-    nested_dataset_cols = [schema_dict[dataset_name][table] for table in dataset_table_names]
-    # unnest for flat list of columns to parse from excel file
-    expected_dataset_cols = [col for table_cols in nested_dataset_cols for col in table_cols]
+    schema_dict = config_dict["schema_definitions"]
```
mmorgantaylor, line 128: why not extract the particular dataset's schema_dict here? rather than having to do `schema_dict[args.dataset_name]` when you feed it into the validate function later on?

mmorgantaylor (commented):
I got through the logic in make_dataset_data_tables.py but haven't gotten through the validate stuff yet!

don't be overwhelmed by the number of comments - many of them are minor/food for thought.

scripts/cfos/dataset_tables_schema.json
```diff
@@ -1,13 +1,213 @@
-{
-    "inph" : {"donor": ["donor_id", "hasOrganismType", "age_at_biopsy", "APOE_atDonorAge", "APOE_value", "follow_up_years",
-                            "has_disease", "has_phenotypic_sex", "MMSE_Biopsy_atDonorAge", "MMSE_Biopsy_value",
-                            "MMSE_Final_atDonorAge", "MMSE_Final_value", "neuropathology"],
-                  "biosample": ["biosample_id", "donor_id", "hasAnatomicalSite", "hasDonorAge", "hasBioSampleType", "biosample_id_of_file"],
-                  "file": ["biosample_id", "DataModality", "donor_id", "summary_file_path", "features_file_path",
-                           "matrix_file_path", "barcode_file_path", "library_id"],
-                  "library": ["library_id", "comment", "donor_id", "UMI_threshold"]
-                },
+{"fields" : {
```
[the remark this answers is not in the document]
KutnerBroadie: The table columns are using field_id. If the field definition needs to be different, I'd argue it isn't actually the same field and the ID at least should be different, and probably the name as well. Let me know if this didn't answer all your questions!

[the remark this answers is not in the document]
KutnerBroadie: Updated the JSON to remove the field_name as we discussed!

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: John requested I add this error when reviewing the code. I'm not sure why it matters whether it's caught or not given the level of error handling elsewhere in the code. It needs to be a fatal error regardless, and the code won't execute without the schema either way.

KutnerBroadie: I think it is in general better practice to have try/except blocks when you can, but in this case I don't think the user will be impacted much either way.

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: I personally like doing it, as we discussed before. :( I like that python does this/allows for this. It may not be necessary everywhere, but since it is auto-suggested to do it by the IDE and I find it helpful, I tend to do it everywhere instead of just some places.

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: Done! Good catch.

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: Yup! I changed it so I could add default values for ease of testing and missed changing it back.

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: In a similar vein, I just updated how use of the validation only flag is handled. I changed it to this, so that if validate_only is set, it overrides any other parameters. This better reflects the behavior intended, I think. 

`validated_dataset = validate_dataset(dataset=dataset_metadata, schema_dict=schema_dict[args.dataset_name], field_dict=column_dict)
    
    if args.validate:
        print("Validation only run complete")
        quit()

    tsv_filenames_list = generate_load_table_files(validated_dataset)
    `

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: If usecols is a list, it requires that ALL the cols be present. That's not what I want. I want it to use it as a filter, not a strict requirement. That's why I changed it. 

skiprows=2 ... as far as I can tell, that's the expected input format. I'm not sure if it's described anywhere, but it is displayed in the example data I provide in the README. 

I can ask Sam to provide a better description of expected input format.

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: I'm only seeing two on my version of the code. Maybe mine is a later version. Can you take a look again once I commit and see if this is resolved?

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: I feel like this is more explicit/safer. The IDE suggests this as well, so I think it's also easier. I'd need to be convinced it's better to not do this for some reason. 

I know this may seem unnecessary, but my brain has trouble keeping track of names/positions/etc. It really helps me to read and write code faster and with fewer mistakes when I use this format of passing in parameters when it's available.

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: Great suggestion, I like this better. :)

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: There was definitely something missing from the validate function, so maybe this was in reference to that? I fixed it. I probably messed it up during my last set of edits. :(

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -143,57 +112,102 @@ def load_excel_input(excel, expected_dataset_cols):
...
     args = parser.parse_args()
+    schema_json = 'dataset_tables_schema.json'
+
+    # TODO: Validate that schema exists; error and quit if not. Store validation code in new requirements or file_paths validation file.
+    # ask John - why write separate validation? Why not just use a Try block? 
+    # https://therenegadecoder.com/code/how-to-check-if-a-file-exists-in-python/#check-if-a-file-exists-with-a-try-block indicates try block is more robust.
 
     # parse schema using dataset name to get list of expected columns
-    expected_dataset_cols, dataset_table_names, schema_dict = get_expected_columns_list(args.schema, args.dataset_name)
+    schema_dict, column_dict = parse_config_file(schema_json=schema_json, dataset_name=args.dataset_name)
```
mmorgantaylor, line 200: I know we talked about this last week, and it does seem that using keywords to specify positional arguments is a matter of stylistic preference. I found this discussion interesting: https://stackoverflow.com/questions/7041752/any-reason-not-to-always-use-keyword-arguments

scripts/cfos/validate.py
```diff
@@ -5,70 +5,256 @@
...
+      return True
+   return False
+
+
+def field_attribute_value_exists(field_id, field_dict, attribute_to_check):
+   if attribute_to_check not in list(field_dict[field_id].keys()):
+      return False
+
+   attribute_value = str(field_dict[field_id][attribute_to_check]).strip()
+   if attribute_value is None or attribute_value == "":
```
mmorgantaylor, line 47: this can be simplified to `return attribute_value is not None and attribute_value != ""`

scripts/cfos/make_dataset_data_tables.py
mmorgantaylor: yeah (and sorry for having multiple threads on this) - especially when there's literally one input and the use is clear from the function name, this really seems like clutter. but I agree it's a stylistic choice and not a hill I want to die on!

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -143,57 +112,102 @@ def load_excel_input(excel, expected_dataset_cols):
...
-    # if validation flag, only validate
-    if args.validate:
-        validated_dataset = validate_dataset(dataset_metadata)
-    else:
-        validated_dataset = validate_dataset(dataset_metadata)
-        dataset_tables_dict = create_dataset_tables_dictionary(validated_dataset, args.dataset_name, dataset_table_names, schema_dict)
-        tsv_filenames_list = generate_load_table_files(dataset_tables_dict)
+    dataset_metadata = load_excel_input(args.excel, list(column_dict.keys()))
+
+    validated_dataset = validate_dataset(dataset=dataset_metadata, field_dict=column_dict, schema_dict=schema_dict)
```
mmorgantaylor, line 204: good call refactoring this out of the if/else logic!

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: Required columns is covered in validation code, not ingest code. I separated it out because I don't think it is good practice to mix together ingest and validation, especially not here. 

It should skip the first two rows - those are headers. The data starts one row 3.

scripts/cfos/README.md
```diff
@@ -56,4 +56,106 @@
...
+	},
+	column_ID2 : {
+		“field_name” : field_name2,
+		“field_type” : field_type2,
+		“other_attributes” : other_values
+	}
+},
+
+“schema_definitions” : {
+	schema_ID1 : {
```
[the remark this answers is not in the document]
KutnerBroadie: This is what was requested originally and I didn't change it. I was told that the different schemas would all have the same fields, just organized different ways.

scripts/cfos/make_dataset_data_tables.py
mmorgantaylor: for usecols: got it, thanks, so basically here as you're reading in the data you're filtering down the column names to the possible expected columns, i.e. dropping any data that's not in your expected list of fields; later you'll check for required ones. makes sense. I'd add a comment to explain the use of a lambda here rather than a list - that's cool functionality but not obvious to the naive reader.

for skiprows: awesome, that's what I was trying to get at with the expected formatting of the excel input. a comment to the effect of "skip 2 rows because the data file has 2 header rows" would be great here.

scripts/cfos/README.md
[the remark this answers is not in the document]
mmorgantaylor: fair enough! are there going to be other datasets included in this release? what are imgl, imperial, and dataset4?

scripts/cfos/README.md
```diff
@@ -27,7 +27,7 @@
             - requires workspace name and workspace project arguments
 
 ##### Usage
-    1. only validate input excel file
+    1. only validate input excel file - if provided, this overrides the tsv creation and "upload" settings
```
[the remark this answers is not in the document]
KutnerBroadie: Correct. Is "only validate input excel file - overrides the tsv creation and "upload" settings" any better? I'm not sure how to make it more clear without drastically changing the comment structure, which I could do, but I'm not sure it's necessary.

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -87,54 +58,52 @@ def generate_load_table_files(dataset_tables_dict):
...
-def load_excel_input(excel, expected_dataset_cols):
+def load_excel_input(excel, allowed_dataset_cols):
     """Read excel file input into a dataframe and validate for required columns."""
 
     try:
-        dataset_metadata_df = pd.read_excel(excel, sheet_name="Sheet1", skiprows=2, usecols=expected_dataset_cols, index_col=None)
-
+        raw_dataset_df = pd.read_excel(excel, sheet_name=None, skiprows=2, index_col=None, usecols=lambda x: x in allowed_dataset_cols)
+        
+    # TODO: Add better file error handling
```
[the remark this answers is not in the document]
KutnerBroadie: Better file error handling just meant more explicit try/catch statements to provide more user-friendly information. This is generally good practice in development instead of letting it fail, but based on what I've heard, that level of error handling isn't necessary here. I can remove the note.

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -143,57 +112,102 @@ def load_excel_input(excel, expected_dataset_cols):
...
-    with open(schema_json) as schema:
-        schema_dict = json.load(schema)
-
-    # get list of tables for a given dataset name
-    dataset_table_names = list(schema_dict[dataset_name].keys())
+    try: 
+        with open(schema_json) as schema:
+            config_dict = json.load(schema)
+    
+    except FileNotFoundError as e:
```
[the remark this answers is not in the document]
KutnerBroadie: I've been taught to do try/catch blocks and handle errors more completely than just letting it fail. I added these so I could go back and add more robust error handling, but it sounds like that isn't necessary for this. I'm still a bit hesitant about removing them without discussing it first because it still seems like good practice to have, but I'm not sure it's something we need here.

scripts/cfos/README.md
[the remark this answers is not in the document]
KutnerBroadie: I'll be working with Sam to figure out what the version of the schema is that we want released with this. I can make sure to remove any unused pieces at that time!

scripts/cfos/validate.py
KutnerBroadie: TODO: Update to 'return attribute_value != ""'

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: Done!

scripts/cfos/make_dataset_data_tables.py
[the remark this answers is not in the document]
KutnerBroadie: Removed it

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: Thanks!

scripts/cfos/validate.py
```diff
@@ -5,70 +5,256 @@
...
-                                                Column("library_id", null_validation),
-                                                Column("UMI_threshold", [IsDtypeValidation(int)]),
-                                                Column("follow_up_years", [IsDtypeValidation(int)]),
-                                                Column("has_disease", [InListValidation(has_disease_options)])
-                                                   #    Column("comment", [None])
-                                            ])
+null_validation = CustomElementValidation(lambda d: d is not np.nan, 'this field cannot be null')
 
+      
+def field_exists_in_df(field_id, df):
```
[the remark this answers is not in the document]
KutnerBroadie: Removed the function!

scripts/cfos/validate.py
KutnerBroadie: This should be updated now!

scripts/cfos/validate.py
```diff
@@ -5,70 +5,256 @@
...
+
+   field_type = field_dict[field_id][type_key]
+
+   if field_type == expected_type:
+      return True
+
+   return False
+
+
+def is_file_path_field_type(field_id, field_dict):
```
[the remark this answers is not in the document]
KutnerBroadie: Updated

scripts/cfos/validate.py
```diff
@@ -5,70 +5,256 @@
...
+   
+   return Schema(columns_logic_list)
+
+
+
+
+# Constants
+
+# Keys
+def get_optional_field_key():
```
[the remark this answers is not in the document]
KutnerBroadie: Done!

scripts/cfos/README.md
```diff
@@ -56,4 +56,106 @@
 ##### Description
     WIP: File containing a dataset (by name) and the dataset's tables, as well as each table's organization structure - which columns are organized into which tables for a given dataset.
 
-    Generated in a json format and can be expanded either by addition additional dataset definisions or appending tables and/or columns to existing definition.
\ No newline at end of file
+    Generated in a json format and can be expanded either by addition additional dataset definisions or appending tables and/or columns to existing definition.
+
+    README along with script in repo (README.md in local git)
```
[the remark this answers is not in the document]
KutnerBroadie: Still working on this - all other requested changes should be addressed though!

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: Pinged @kilonzi on slack asking about this again. I think it may be fine to remove this/not include it.

scripts/cfos/README.md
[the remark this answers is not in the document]
KutnerBroadie: Should be addressed now

scripts/cfos/make_dataset_data_tables.py
kilonzi: Good catch here @mmorgantaylor, this is probably outdated but we can remove try/catch. A more general way to look at it is through [LBYL](https://docs.python.org/3.5/glossary.html#term-lbyl) and [EAFP](https://docs.python.org/3.5/glossary.html#term-eafp). We removing it here because it provided no additional benefit, other than fail and we do not anticipate to handle the exception in any way.

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -87,113 +58,155 @@ def generate_load_table_files(dataset_tables_dict):
...
-    except ValueError as e:
-        # if the value error is about missing columns in input file
-        if len(e.args) > 0 and e.args[0].startswith("Usecols do not match columns, columns expected but not found"):
-            missing_cols = e.args[0].split(":")[1]
-            print(f"Failed: Input excel file has the following missing columns that are required: {missing_cols}")
-        else:
-            print(e)
-        exit()
-    
+    raw_dataset_df = pd.read_excel(excel, sheet_name=None, skiprows=skiprows, index_col=None, usecols=lambda x: x in allowed_dataset_cols)
```
mmorgantaylor, line 110: as discussed on Slack: consider adding an "FYI" comment here to notify the user of any columns being dropped at ingest, due to not being in the schema

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -31,47 +33,16 @@ def upload_dataset_table_to_workspace(tsv_filenames_list, workspace_name, worksp
...
 
     # for each table in dictionary
-    for table_name in dataset_tables_dict:
+    for table_name in list(dataset_tables_dict.keys()):
         table_df = dataset_tables_dict[table_name]
-        if table_name == "file":
-            table_df = modify_file_table_dataframe(table_df)
+            
         # change table name column to new name --> entity:table_name_id and set as index
         table_df = table_df.rename(columns={f"{table_name}_id": f"entity:{table_name}_id"})
```
mmorgantaylor, line 47: as discussed on Slack: rather than requiring the schema to have a `{table_name}_id` field, you could do what TDR does and have the schema specify which field should be used as the table's primary key, and then copy it to a new field called `{table_name}_id` (leaving the original field named as-is - yes this duplicates data but (a) consistent logic with TDR import to Terra and (b) avoids confusion if the end user is expecting that field named just so.

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: I think I've removed this and the comments now.

scripts/cfos/make_dataset_data_tables.py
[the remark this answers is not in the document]
KutnerBroadie: Also, not sure why I can't reply to the questions about the column/row skipping:

I added an optional parameter so that users can set the number of rows to skip themselves, but it defaults to 2. I'm pretty sure (based on the files Sam has provided) that the first 2 rows are just extra info and the column headers start on the 3rd row. I'll make sure to check with Sam and confirm that this is still the case.

regarding usecols: I separated validation from reading the files. It will read the file just fine if required columns are missing, but it will fail during validation if it is missing a column that was specified in the schema as belonging to that table.

scripts/cfos/README.md
```diff
@@ -50,10 +53,110 @@
...
+
+The JSON is split into two sections. One section lists and describes all the columns of data expected in the input. By default, these columns are treated as required for successful data ingest. Another section defines various schemas which indicate the expected organization of the files, both in the input file and upon upload to the database.
+
+What are schemas? 
+Schemas are different ways of organizing the columns of data. This will determine how many tables will be created to store the data in Terra, and which columns appear in which tables. All columns specified in the schema being used must be present in the input file for successful data upload.
+
+Which schema to use is specified as an input to the upload script at run time.
+
+How do you define columns?
+ Columns of data are defined in the “Fields” section of the JSON document. Each column must have a unique column_ID, which is used to identify the column in the input document. The column_ID is the value under which the rest of the column information is stored in the JSON dictionary.
```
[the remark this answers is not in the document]
KutnerBroadie: Column_ID and column name are functionally the same now. Can you let me know which one you think is best and I can run through and try to make the terms consistent?

scripts/cfos/README.md
```diff
@@ -50,10 +53,110 @@
...
+
+
+Each field has the following attributes. Attributes that are “type-specific” are listed under the corresponding field “data type”. 
+
+
+Field:
+column_ID
+field_name
+field_type
+value_required (optional - defaults to “yes?”)
```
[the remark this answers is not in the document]
KutnerBroadie: Good catch - pretty sure this is because I made this value call but wanted my reviewer to confirm that's what they would expect as well. Then I forgot about it. :) 

Does this match your expectations? I'll go ahead and remove the "?" since it's served its purpose.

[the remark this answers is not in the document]
KutnerBroadie: I actually had this as "defaults to no" in the code, so I updated this to be consistent with it for now, but can change both if we think it should be the opposite.

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -87,113 +58,145 @@ def generate_load_table_files(dataset_tables_dict):
...
+            field_dict=field_dict, 
+            fields_to_validate_list=schema_dict[table], 
+            data_df=dataset[table]
+            )
+        # get indices of the failed rows
+        errors_index_rows = [e.row for e in errors]
+
+        # if any errors, print error message and exit
+        if errors_index_rows:
+            print(f"Failed: Dataset validation errors found in table {table}. Please retry after correcting errors listed in validation_errors.csv.")
```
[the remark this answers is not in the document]
KutnerBroadie: got it

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -87,113 +58,145 @@ def generate_load_table_files(dataset_tables_dict):
...
+            pd.DataFrame({'validation_error':errors}).to_csv('validation_errors.csv')
+            exit()
+
+        # TODO: determine how to handle drops of failed rows/columns to get cleaned data
```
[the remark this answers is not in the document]
KutnerBroadie: According to [this](https://pandas.pydata.org/docs/reference/api/pandas.read_excel.html) documentation, it looks like pandas.read_excel reads multiple sheets into a dictionary of dataframes. 

The validation code loops over each "sheet" (sheet = table, they're the same thing) and passes in the corresponding dataframe to validate.py for validation.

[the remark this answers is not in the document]
KutnerBroadie: Regarding dataset being the same as validated_dataset, I thought so too, but I think they were treated differently in the initial code and I wasn't sure if it'd impact later code. 

In the previous code, since it was evaluating each table separately, it needed to join all the tables together as they were marked valid. Here, we already have a dict of dataframes, and it may not be necessary anymore. 

I'm not sure if there's code downstream of this that could be consolidated too, I tried to keep my changes relatively contained to the validation logic itself and the reading of the excel.

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -87,113 +58,145 @@ def generate_load_table_files(dataset_tables_dict):
...
-    parser.add_argument('-v', '--validate', required=False, action='store_true', help='Set parameter to run only validation on input excel file.')
-    parser.add_argument('-u', '--upload', required=False, action='store_true', help='Set parameter to upload data table files to workspace. default = no upload')
+    # Goal: allow qualified people to make easy updates to schemas, but don't encourage random users to play with it.
+
+    parser.add_argument(
+        '-d', 
+        '--dataset_name', 
+        required=False,
+        type=str, 
+        help='Dataset type to apply validations and data table structure. ex. proteomics, atac-seq, transcriptomics, singe-cell'
```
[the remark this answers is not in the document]
KutnerBroadie: Added it, not sure how to make this more clear while remaining succinct. If you have any suggestions, I'd appreciate.

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: per slack:
So I can't figure out an elegant solution for notifying the user that some columns were dropped. The read_excel API doesn't seem to have an option.
I'm not sure if there's a way to rope that in as a side effect of the lambda function that's evaluating if the column should be included or not.
I could just print the data frames as they exist after import, allowing the user to look over them and see if anything is missing that they expect should be there.
Would that be a reasonable solution if we can't figure something better out?

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: Updated! Ran the code and it duplicated and designated the appropriate columns as index columns, renamed to entity:tablename_id

scripts/cfos/tests/README.md
```diff
@@ -0,0 +1,18 @@
+## Test schema file setup ##
+You can either:
+- replace the contents of 'scripts/cfos/dataset_tables_schema.json' file with the contents of the 'scripts/cfos/tests/test_schema.json' file temporarily OR
+- change the code in make_dataset_data_tables.py so that schema_json points to the test schema file instead of the real one
```
mmorgantaylor, line 4: can you make a "schema_file" optional input that defaults to the production one, rather than requiring the user to change the code or edit files?

scripts/cfos/tests/README.md
```diff
@@ -0,0 +1,18 @@
...
+## How to Run ##
+### test fail ### 
+python3 make_dataset_data_tables.py -d schema1 -x tests/CFoS_Template_Test_Schema1_Fail.xlsx -p broad-cfos-data-platform1 -w cFOS_automation_testing
+
+### test pass ###
+python3 make_dataset_data_tables.py -d schema1 -x tests/CFoS_Template_Test_Schema1_Pass.xlsx -p broad-cfos-data-platform1 -w cFOS_automation_testing
+
+
+## Validating outcomes ##
+Check the validation_erros.csv and make sure expected outcomes are met.
```
mmorgantaylor, line 15: the expected outcome should ideally be shown here in the README rather than in the data file

scripts/cfos/README.md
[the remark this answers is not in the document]
mmorgantaylor: personally I would say column_name, since that's more user-friendly in my opinion, but I'd trust your judgment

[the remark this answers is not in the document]
KutnerBroadie: Updated to use column_name instead of column_ID

scripts/cfos/README.md
```diff
@@ -50,10 +53,110 @@
...
+ Columns of data are defined in the “Fields” section of the JSON document. Each column must have a unique column_ID, which is used to identify the column in the input document. The column_ID is the value under which the rest of the column information is stored in the JSON dictionary. 
+
+Note: if you have the same field in multiple tables, the same validation will apply to both, unless they are named something unique. AKA: if "filename" in one table should be alphanumeric and in another table should be an integer, you need to name the fields "tablex_filename" and tabley_filename" to apply different validation.
+
+
+Each field has the following attributes. Attributes that are “type-specific” are listed under the corresponding field “data type”. 
+
+
+Field:
+column_ID
```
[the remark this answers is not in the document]
KutnerBroadie: do you mean changing column_ID to column_name? Is this different from the above suggestion of using column_name consistently in the document?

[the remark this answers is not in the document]
mmorgantaylor: yeah, this would be a logical change going along with the previous one to generally refactor this key name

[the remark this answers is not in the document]
mmorgantaylor: oh or I guess the consistent thing (looking at your schema files) is to remove this field altogether.

scripts/cfos/README.md
```diff
@@ -50,10 +53,110 @@
...
+pattern_to_match (optional)?
+file_path
+pattern_to_match (optional)?
+
+Format for JSON:
+{“fields” : {
+	column_ID1 : {
+		“field_name” : field_name1,
+		“field_type” : field_type1,
+		“other_attributes” : other_values
```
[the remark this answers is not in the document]
KutnerBroadie: Updated to say <attribute_key> : <attribute_value>

scripts/cfos/README.md
```diff
@@ -50,10 +53,110 @@
...
+	column_ID2 : {
+		“field_name” : field_name2,
+		“field_type” : field_type2,
+		“other_attributes” : other_values
+	}
+},
+
+“schema_definitions” : {
+	schema_ID1 : {
+		table_ID1 : {
```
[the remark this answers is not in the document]
KutnerBroadie: Updated to also reflect the change with the primary key designation

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -87,113 +58,145 @@ def generate_load_table_files(dataset_tables_dict):
...
+            print(dataset.keys())
+            exit()
+
+        errors = dynamically_validate_df(
+            field_dict=field_dict, 
+            fields_to_validate_list=schema_dict[table], 
+            data_df=dataset[table]
+            )
+        # get indices of the failed rows
+        errors_index_rows = [e.row for e in errors]
```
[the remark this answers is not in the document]
KutnerBroadie: I was just leaving this as it was - it looks like it works (and makes more sense) to just use `if errors:`

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -87,113 +58,145 @@ def generate_load_table_files(dataset_tables_dict):
...
 
+    return False
 
-def load_excel_input(excel, expected_dataset_cols):
-    """Read excel file input into a dataframe and validate for required columns."""
 
-    try:
-        dataset_metadata_df = pd.read_excel(excel, sheet_name="Sheet1", skiprows=2, usecols=expected_dataset_cols, index_col=None)
+def load_excel_input(excel, allowed_dataset_cols):
+    """
```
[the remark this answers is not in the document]
KutnerBroadie: Added: Returns: dictionary of data frames where each sheet is a separate dataframe in the dictionary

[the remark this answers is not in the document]
mmorgantaylor: ah thanks! I think it would be helpful to add a comment as well after line 122 with this info, which was not obvious to me
When `sheet_name=None` is specified, all sheets are returned, as a {sheet_name:dataframe} dictionary.

scripts/cfos/make_dataset_data_tables.py
mmorgantaylor: What if you just read in all the data here (i.e. remove usecols limitation) and in the validation you highlight the extra fields present?

Do you know from the user whether they'd prefer to keep extra data or drop it?

scripts/cfos/tests/README.md
KutnerBroadie: Instead, I added a "test" parameter that will run this using the test schema.

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -87,113 +60,174 @@ def generate_load_table_files(dataset_tables_dict):
...
+            a list enforces that exactly those columns must exist in the data frame, and we don't want to
+            enforce the schema until we're checking it during validation. This should solely be data import.
+    """
 
+    raw_dataset_df = pd.read_excel(excel, sheet_name=None, skiprows=skiprows, index_col=None, usecols=lambda x: x in allowed_dataset_cols)
 
-def load_excel_input(excel, expected_dataset_cols):
-    """Read excel file input into a dataframe and validate for required columns."""
+    print("Success: Excel file has been loaded into a dataframe.")
+    return raw_dataset_df
```
mmorgantaylor, line 128: I'd rename raw_dataset_df to something like raw_dataset_dict since it's a dictionary, not a dataframe - this was the source of some of my confusion

scripts/cfos/README.md
```diff
@@ -30,6 +30,10 @@
         2. upload flag - pushes generated load table tsv files to Terra workspace
             - requires workspace name and workspace project arguments
 
+	Output: 
+		- If errors occur, generates and populates validation_errors.csv with a list of all errors found. 
+		- If successful, generates a separate .tsv file for each table in the schema
```
mmorgantaylor, line 35: nice! to be clear, if successful, it will also upload those tsvs to Terra, right? I get why you didn't say that here since this is the outputs section but it could be helpful to mention in a parenthetical that those are the tsvs that also get ingested to the Terra data tables.

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: I talked to the user about handling extra data, but upon discussing with John it was determined to be scope creep and out of scope for this. 

I'm just not sure that this warrants getting rid of the out-of-the-box way of filtering columns on ingest for the read_excel API. I added a notification that pops up when that code is run to notify the user that any columns not present in the schema will be ignored.

scripts/cfos/make_dataset_data_tables.py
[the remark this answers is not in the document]
KutnerBroadie: What does adding the comment add that the docstring doesn't? Calling out that sheet_name=None is what causes this? Would I just be able to add that to the existing docstring instead of adding another comment elsewhere?

scripts/cfos/tests/README.md
KutnerBroadie: I think I addressed this with the examples of validation and what error messages would look like.

scripts/cfos/tests/README.md
```diff
@@ -1,18 +1,24 @@
...
-python3 make_dataset_data_tables.py -d schema1 -x tests/CFoS_Template_Test_Schema1_Pass.xlsx -p broad-cfos-data-platform1 -w cFOS_automation_testing
+python3 make_dataset_data_tables.py -d schema1 -x tests/CFoS_Template_Test_Schema1_Pass.xlsx -p broad-cfos-data-platform1 -w cFOS_automation_testing -t True
 
 
 ## Validating outcomes ##
 Check the validation_erros.csv and make sure expected outcomes are met. 
 General expected outcomes are explained above the column ID in the Fail.xlsx file. 
 
+The code can fail in the following ways:
+- **primary key column contains non-unique values**: 'validation_error':{row: 6, column: "id_unique"}: "X1234B" contains values that are not unique
```
mmorgantaylor, line 17: is "X1234B" a value or a column name? I think it's a value (the message suggests that the column name is "id_unique") - in which case I find this wording confusing - how can a single value contain values that are not unique? it points to row 6, but again wouldn't this suggest that there are multiple rows containing duplicate values?

scripts/cfos/tests/README.md
```diff
@@ -1,18 +1,24 @@
...
+- **value doesn't match an expected pattern** : 'validation_error':{row: 6, column: "file_path_explicit_pattern"}: "gs://456" does not match the pattern "^fs://"
```
mmorgantaylor, line 18: minor, but it's confusing to me that this example uses a different string match than you'd likely expect - like I'd expect the pattern to match to be "^gs://" and the example non-matching string to be "fs://456". for a second i thought i was misunderstanding the pattern formatting.

scripts/cfos/tests/README.md
```diff
@@ -1,18 +1,24 @@
...
+- **value is an unexpected type**. If "number", it must be a float. If "integer", it must be an integer: 'validation_error':The column integer_only_field has a dtype of object which is not a subclass of the required type <class 'int'>
```
mmorgantaylor, line 19: is there a reason this one is formatted differently from the others? the others have a row and column identification before the message (which is useful for the user to identify which data row the offending value came from.

scripts/cfos/tests/README.md
```diff
@@ -0,0 +1,24 @@
...
+## How to Run ##
+### test fail ### 
+python3 make_dataset_data_tables.py -d schema1 -x tests/CFoS_Template_Test_Schema1_Fail.xlsx -p broad-cfos-data-platform1 -w cFOS_automation_testing -t True
+
+### test pass ###
+python3 make_dataset_data_tables.py -d schema1 -x tests/CFoS_Template_Test_Schema1_Pass.xlsx -p broad-cfos-data-platform1 -w cFOS_automation_testing -t True
+
+
+## Validating outcomes ##
+Check the validation_erros.csv and make sure expected outcomes are met.
```
mmorgantaylor, line 13: minor: typo "validation_errors.csv"

scripts/cfos/tests/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+## How to Run ##
+### test fail ### 
+python3 make_dataset_data_tables.py -d schema1 -x tests/CFoS_Template_Test_Schema1_Fail.xlsx -p broad-cfos-data-platform1 -w cFOS_automation_testing -t True
+
+### test pass ###
+python3 make_dataset_data_tables.py -d schema1 -x tests/CFoS_Template_Test_Schema1_Pass.xlsx -p broad-cfos-data-platform1 -w cFOS_automation_testing -t True
+
+
+## Validating outcomes ##
```
mmorgantaylor, line 12: I would move this entire section to the top-level README - I think this is useful for the user to see. I'd also love to see the literal contents of `validation_errors.csv` when run on the failing test, but if it's similar enough to the examples you gave below, I think that's ok.

scripts/cfos/validate.py
```diff
@@ -5,70 +5,188 @@
...
-                                                Column("matrix_file_path", [MatchesPatternValidation("gs://")]),
-                                                Column("barcode_file_path", [MatchesPatternValidation("gs://")]),
-                                                Column("biosample_id_of_file", null_validation),
-                                                Column("library_id", null_validation),
-                                                Column("UMI_threshold", [IsDtypeValidation(int)]),
-                                                Column("follow_up_years", [IsDtypeValidation(int)]),
-                                                Column("has_disease", [InListValidation(has_disease_options)])
-                                                   #    Column("comment", [None])
-                                            ])
+#null_validation = CustomElementValidation(lambda d: d is not np.nan, 'this field cannot be null')
```
mmorgantaylor, line 8: cleanup: remove commented line

scripts/cfos/validate.py
```diff
@@ -5,70 +5,188 @@
...
+def field_attribute_value_exists(field_id, field_dict, attribute_to_check):
+   if attribute_to_check not in field_dict[field_id]:
+      return False
 
+   attribute_value = str(field_dict[field_id][attribute_to_check]).strip()
+   return attribute_value is not None and attribute_value != ""
 
 
+def check_field_type(field_id, field_dict, expected_type):
+   type_key = FIELD_TYPE_KEY
```
mmorgantaylor, line 36: rather than duplicating this variable, you can modify the following line to be `if FIELD_TYPE_KEY not in field_dict[field_id]:`

and similarly row 40 to `field_type = field_dict[field_id][FIELD_TYPE_KEY]`

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -87,113 +60,185 @@ def generate_load_table_files(dataset_tables_dict):
...
+    validated_dataset = validate_dataset(dataset=dataset_metadata, field_dict=column_dict, schema_dict=schema_dict, primary_key_dict=primary_keys_dict)
+    
+    if args.validate_only:
+        print("Validation only run complete")
+        quit()
+
+    tsv_filenames_list = generate_load_table_files(validated_dataset, primary_keys_dict)
     
     # if upload flag, upload generated tsv files to Terra workspace
     if args.upload and args.workspace and args.project:
```
mmorgantaylor, line 244: I know this isn't part of your changes, but just saw this - will this fail (to upload to Terra) silently if the user includes the --upload flag but forgets to include both the --workspace and --project inputs? might be worth a quick edit to notify the user in that case.

scripts/cfos/make_dataset_data_tables.py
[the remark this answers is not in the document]
mmorgantaylor: yeah, exactly, that it's `sheet_name=None` that does this. and yeah I was suggesting adding it to your docstring after your other explanations of the parameters you pass to read_excel.

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -87,113 +60,185 @@ def generate_load_table_files(dataset_tables_dict):
...
-    print(f"Success: Dataset has been validated.")
-    return dataset
+        Returns: dictionary of data frames where each sheet is a separate dataframe in the dictionary
+    """
 
+    raw_dataset_dict = pd.read_excel(excel, sheet_name=None, skiprows=skiprows, index_col=None, usecols=lambda x: x in allowed_dataset_cols)
 
-def load_excel_input(excel, expected_dataset_cols):
-    """Read excel file input into a dataframe and validate for required columns."""
+    print("Success: Excel file has been loaded into a dataframe. Note that any columns that are not defined in the schema will be ignored.")
```
mmorgantaylor, line 128: UX suggestion: tell the user how many rows and columns of data were read in, as a sanity check.

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -87,113 +58,145 @@ def generate_load_table_files(dataset_tables_dict):
...
-    """Validate the full dataset metadata with Schema validator."""
-
-    # capture errors from the validation against the pre-defined validations per column
-    errors = schema_validator.validate(dataset, columns=schema_validator.get_column_names())
-    # get indices of the failed rows
-    errors_index_rows = [e.row for e in errors]
+def validate_dataset(dataset, schema_dict, field_dict):
+    validated_dataset = {}
+
+    for table in list(schema_dict.keys()):
```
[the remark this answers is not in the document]
mmorgantaylor: (did you accidentally resolve this?)

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: I'm not sure this is worth doing. I'd have to loop over the dictionary for each data frame and print the name of the table and the size of the data, and even then, they won't be able to tell what's missing if something is missing. 

Would it make more sense to just print the dictionary of data frames?

scripts/cfos/make_dataset_data_tables.py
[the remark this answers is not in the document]
KutnerBroadie: added

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: found this. pretty nifty! https://stackoverflow.com/questions/25626109/python-argparse-conditionally-required-arguments

scripts/cfos/tests/README.md
KutnerBroadie: The examples are excerpts of the `validation_errors.csv`. I think showing the whole file would be overwhelming unnecessarily unless someone is actively trying to debug their data.

KutnerBroadie: Should I just get rid of the tests README then? ANd move it all to the main one?

scripts/cfos/tests/README.md
KutnerBroadie: This is the value. It is indicating which values in the column are not unique. In the actual version it'd write that for every non-unique value in the column, but I don't think that's necessary for this example. 

This is the output for the pandas_schema is_distinct validation.

scripts/cfos/tests/README.md
KutnerBroadie: It doesn't really matter either way - in this case I pulled the error for the case where I over-wrote the default pattern matching and was testing to make sure it applied that instead of the default. Therefore "^gs" should fail because it expects something different. 

I went ahead and changed the values - the purpose of this isn't that this value will always be the required pattern to match, but rather that this is the format of the error message when a pattern is not matched.

scripts/cfos/tests/README.md
KutnerBroadie: I'm not sure why this is different formatting. They're all the default error messages from Pandas_schema.

mmorgantaylor (approved):
let's ship it

PR #128, opened 2023-02-17
FE-4 update workspace_cleanup with docker

bahill:
[FE-4](https://broadworkbench.atlassian.net/browse/FE-4)
Sam was having issues with scripts/workspace_cleanup/delete_workspaces.py. I verified that this works in the base horsefish Docker image, but wanted to create a tagged version, which required updating the image and adding a helper script to build.
NB I do not currently have push rights to the [horsefish dockerhub,](https://hub.docker.com/r/broadinstitute/horsefish) but I have tested the newly built image locally. I will push on Tuesday as soon as I'm given access.

[FE-4]: https://broadworkbench.atlassian.net/browse/FE-4?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/anvil_tdr_ingest/clone_anvil_workspace.py
```diff
@@ -5,8 +5,9 @@
 
 import argparse
 import json
-import pandas as pd
-import requests
+# not used some commenting out for now
```
mmorgantaylor, line 8: please delete these commented out lines if they're not used!

scripts/workspace_cleanup/README.md
```diff
@@ -8,6 +8,15 @@ The scripts in this directory can be used to delete Terra workspaces in bulk. Pr
 
 These scripts use the Terra APIs, so they can only delete workspaces that the running user has access to. If your goal is to view as many workspaces as possible in the project, it's best to make sure the person running the script is a project owner on the Terra project. Note that any workspace covered by an Authorization Domain (AD) will not be able to be deleted (or have cost information returned) by these scripts unless the person running the script is in the workspace's AD.
 
+## Docker image & run command
+To run the scripts in this directory, you can use the Docker image `broadinstitute/horsefish:latest`. \
```
mmorgantaylor, line 12: do we want to use `latest` or a specific tag?

update_docker_image.sh
```diff
@@ -0,0 +1,10 @@
+set -e
+
+# DEVELOPER: update this field anytime you make a new docker image
+VERSION="2.0"
```
mmorgantaylor, line 4: would it make sense to force the developer to input the version number? I worry about folks not realizing they need to update it. 

in other places I put this version in the code itself and have this script pull it out - admittedly also not a great solution but more likely for the developer to see it than here, in my opinion.

mmorgantaylor: oh I see this is for the top-level Dockerfile. I thought we were going to remove this and have a docker image per script. but this works, too. 

in this case, since all scripts are included in this image, I agree it makes sense to have this here, but I'd want to do something more to force the developer to increment the version. 

the nicest thing would be a prompt that checks for existing builds with that tag and requires the developer/user to confirm that they want to overwrite the existing build with that tag (which will likely happen during development, so we don't want to auto-increment). what do you think?

scripts/anvil_tdr_ingest/clone_anvil_workspace.py
bahill: Sorry forgot about that cleanup

bahill: cleaned up now.

scripts/workspace_cleanup/README.md
bahill: I meant to use a specific tag - I was actually making those updates while you were reviewing =)

update_docker_image.sh
bahill: Yeah - I saw that you'd done that for twistcap, but as you said, it was top-level, so there wasn't specific code to pull from. I could instead just tag this as workspace_cleanup and move the script there. I think I got sidetracked because there wasn't a script for the top-level Docker.
I think a prompt is good. That is what my previous team did for this sort of thing. 
AND/OR should I make a workspace_cleanup-specific image? (for now, it'll be just a copy of the top level, but may diverge one day - and then I wouldn't copy in the whole scripts dir, which felt like overkill...)

mmorgantaylor: I'd say do the minimum necessary for this PR - so don't split it out now if there's no need. Can always be done in future.

update_docker_image.sh
```diff
@@ -0,0 +1,19 @@
...
+
+echo "using build tag: ${BUILD_TAG}"
+
+if docker manifest inspect $BUILD_TAG > /dev/null ;
+then
+    echo "Docker image already exists. Do you want to overwrite this tag?."
+    read -p "Press [Enter] to overwrite or [Ctrl+C] to exit and change the version number."
+fi
+
+echo "Docker image does not exist. Building new image."
```
mmorgantaylor, line 16: don't you want this in an else statement? looks like you could get here even after confirming you want to overwrite the existing tag, right?

update_docker_image.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# DEVELOPER: update this field anytime you make a new docker image
+VERSION="2.1"
+BUILD_TAG="broadinstitute/horsefish:v${VERSION}"
+
+echo "using build tag: ${BUILD_TAG}"
+
+if docker manifest inspect $BUILD_TAG > /dev/null ;
+then
+    echo "Docker image already exists. Do you want to overwrite this tag?."
```
mmorgantaylor, line 12: would be nice to print the name of the tag here (I know you just echoed it above - but still) - could change this to 
`echo "Docker image with tag ${BUILD_TAG} already exists. Do you want to overwrite this tag?"`

mmorgantaylor (approved):
just a couple minor suggestions about echo statements in the bash script, otherwise looks good to me!

update_docker_image.sh
bahill: I thought so too (I've never used bash for conditional logic) but it seems the else is implicit. (https://www.thegeekstuff.com/2010/06/bash-if-statement-examples/)
certainly is less typing anyway
BUT my echo statement is from when I had started to write the else, so that's confusing. I'll change the wording there.

update_docker_image.sh
bahill: HA - I hadn't even read this yet and added that in. Great minds!

update_docker_image.sh
mmorgantaylor: hmm, I'm not sure I follow about the else being implicit - that would suggest that you'll only get to this echo statement if the image doesn't already exist (the condition of the if statement). but you'll get here even if the image exists but the user has pressed Enter, right?

update_docker_image.sh
mmorgantaylor: did you add it in? i don't see it. not a huge deal :shrug:

PR #130, opened 2023-03-10
DFE-805 cfos data tables demo updates

KutnerBroadie:
Adding some features requested during a demo to CFoS team:
- update error log to generate with a time stamp so it isn't overwritten
- make error log a txt file
- print out column names instead of entire table contents

Also fixing a bug where upload fails if primary key name and table name are the same, due to creation of a duplicate key column on upload. 

See original pr for more details: https://github.com/broadinstitute/horsefish/pull/127


mmorgantaylor:
hey @KutnerBroadie since you haven't yet merged your first PR, we can't see your changes here. Could you either merge that PR and then rebase this one, or make this PR against your other branch?

PR #131, opened 2023-03-10
Rek cfos demo updates

KutnerBroadie:
Adding some features requested during a demo to CFoS team:

update error log to generate with a time stamp so it isn't overwritten
make error log a txt file
print out column names instead of entire table contents
Also fixing a bug where upload fails if primary key name and table name are the same, due to creation of a duplicate key column on upload.

See original pr for more details: https://github.com/broadinstitute/horsefish/pull/127


scripts/cfos/make_dataset_data_tables.py
```diff
@@ -129,7 +134,10 @@ def load_excel_input(excel, allowed_dataset_cols, allowed_dataset_tables, skipro
     raw_dataset_dict = pd.read_excel(excel, sheet_name=None, skiprows=skiprows, index_col=None, usecols=lambda x: x in allowed_dataset_cols)
     processed_dataset_dict = {k:v for (k,v) in raw_dataset_dict.items() if k in allowed_dataset_tables}
     print("Success: Excel file has been loaded into a dataframe. Note that any columns that are not defined in the schema will be ignored.")
-    print(processed_dataset_dict)
+    for key, value in processed_dataset_dict.items():
+        print("table: " + key)
```
mmorgantaylor, line 138: do you want to print the table name once for each column? I think you might want to move this line to before this for-loop.

also a good place to use f-strings here (`f"table: {key}"`) and in the next row

scripts/cfos/make_dataset_data_tables.py
```diff
@@ -129,7 +134,10 @@ def load_excel_input(excel, allowed_dataset_cols, allowed_dataset_tables, skipro
...
+        print(key + " columns: " + str(value))
+        print("\n")
```
mmorgantaylor, line 140: I'd also put this newline outside of the for-loop so you don't have a break between every column name, but rather between every table. but that's just personal preference - if you intended to have a space between every column name, feel free to ignore me

KutnerBroadie: d'oh you're right! thanks!

mmorgantaylor (approved):
looks good! one suggestion about how you print out the column names etc, otherwise 👍

scripts/cfos/make_dataset_data_tables.py
KutnerBroadie: Actually this for loop needed to be changed in a slightly different way. The key is the table name, the value is the dataframe itself which has the columns. But I wasn't grabbing the columns correctly and it was getting wonky.

PR #138, opened 2023-06-02
Update create_md5_object WDL to create backup copy, tmp copy, and compare file sizes before overwriting

schaluva:
WDL and python script that generate an object md5 for objects that were uploaded to GCS with some method that does not inherently generate an md5 - like with parallel composite uploads.

Summary:
1. If user provides a `backup_dir`, make a copy of the original object in the user defined location - same name as original object.
2. Make sure that the original object and copy in the `backup_dir` are the same size in bytes.
3. Always generate a tmp copy of the original object - same location as original object but object name is appended with ".tmp". This step happens with `gsutil cp -D` daisy-chain mode to ensure that an md5 is generated. This step is also the rate limiting step.
4. Make sure that the original object and tmp copy are the same size in bytes.
5. Create a copy of the tmp object with the same name as original object - essentially overwriting the original object with no md5 with the tmp copy that does have an md5.
6. Delete the tmp object.

All the copy steps above are performed with the google cloud storage client which make copies faster except for the creation of the tmp object which we do with the `gsutil cp -D` which generates the md5 version of the object.

Resources:
1. Dockstore WDL: https://dockstore.org/workflows/github.com/broadinstitute/horsefish/create_object_md5:create_md5_wdl_v2
2. Test Terra Workspace: https://app.terra.bio/#workspaces/broad-firecloud-dsde/anvil_md5_generation_sushmac_development


scripts/object_hashing/create_object_md5.py
```diff
@@ -0,0 +1,121 @@
...
+    logging.info(f"dest object size: {obj_2_bytes} bytes with md5 {obj_2_md5}")
+
+    if obj_1_bytes != obj_2_bytes:
+        raise ValueError(f"Copy of object failed - {object_1_blob} and {object_2_blob} do not have the same file size. \
+            This is likely a transient failure. Please submit workflow again.")
+    
+    logging.info(f"Copy of object complete - {object_1_blob} and {object_2_blob} have the same file size. \n")
+
+
+def get_file_size(bucket_name, object_name, project_id=None):
```
snovod, line 87: I think you end up always calling this twice for the file, once when comparing objects to make sure they match and once when getting md5.

scripts/object_hashing/create_object_md5.py
```diff
@@ -0,0 +1,121 @@
...
+    create_md5_object(original_bucket_name, original_blob_name, tmp_blob_name, project_id)
+
+
+def create_backup_object(backup_directory, original_bucket_name, original_blob_name, project_id=None):
+    """Make copy of object in provided location."""
+
+    # TODO: check for slash at end
+    logging.info("Backup directory has been provided.")
+    backup_bucket_name = backup_directory.split("/")[:-1][2] # fc-
+    backup_blob_name = "/".join(backup_directory.split("/")[3:]) + "".join(original_blob_name.split("/")[-1:])
```
snovod, line 63: You can use `os.path.join()`. This will make it so it doesn't matter if `/` is there or not

scripts/object_hashing/create_object_md5.py
```diff
@@ -0,0 +1,121 @@
...
+    
+    original_bucket_name = args.original_object.split("/")[:-1][2] # fc-***
+    original_blob_name = "/".join(args.original_object.split("/")[3:]) # dir/filename.txt
+
+    # if the user provides a backup directory
+    if args.backup_dir:
+        create_backup_object(args.backup_dir, original_bucket_name, original_blob_name, args.requester_pays_project)
+    
+    # always create tmp object
+    create_tmp_object(original_bucket_name, original_blob_name, args.requester_pays_project)
```
snovod, line 26: I think might be better to separate the get md5 part and not have the be called from create_tmp_object. To me that is a little clearer. Also 100% preference and works great as it is.

snovod (approved):
Looks good! Few comments but nothing actually needs to change.

snovod (approved):

ncalvanese1 (approved):
Logic looks good to me. Left one comment on something I wasn't sure on

pipelines/create_object_md5.wdl
```diff
@@ -2,80 +2,48 @@ version 1.0
...
-        gsutil -u anvil-datastorage mv -c -L create_md5_log.csv "~{tmp_object_path}" "~{src_object_path}"
-    }
+    command <<<
 
-    runtime {
-        docker: "gcr.io/google.com/cloudsdktool/cloud-sdk:305.0.0"
-        disks: "local-disk " + (disk_size + 2) + " SSD"
-        memory: select_first([memory, 2]) + " GB"
-    }
+        python3 /scripts/create_object_md5.py -o ~{original_object} ~{"-d=" + backup_object_dir} ~{"-r=" + requester_pays_project}
```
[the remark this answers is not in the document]
schaluva: nope not ignorance! 
the create_object_md5.py script lives at the `scripts/object_hashing` home within the repo but when i built the docker image to home this script, I used a Dockerfile and in the Dockerfile I copied the python script to the `/scripts` directory _within_ the docker image. aka in the Dockerfile you can see on line 6 that we copy the python script to the `/scripts` folder and so from the WDL we have to call it from the location at which it lives in the image.

PR #139, opened 2023-06-09
[AT-412] Multi-bucket rsync script

KutnerBroadie:
Script to sync between a source google bucket (aka staging) and a destination google bucket (aka production). 

Takes in a tab-delimited file with a header and two columns: 'Destination directory' and 'Source directory'.

It runs gsutil rsync for each pair of folders, copying all changed files from destination to source.

I'll be writing up a FE ticket to create a ops portal workflow that does this (so PjMs can take over this part of the ingest workflow completely), but this is a prototype/can be used ad hoc in the meantime.

### DIFF
--- a/scripts/anvil_tools/gsutil_bucket_sync.py
+++ b/scripts/anvil_tools/gsutil_bucket_sync.py
@@ -21,13 +21,8 @@ def parse_args():
     return parser.parse_args()
 
 
-def get_paths_df(input_file):
-    """Creates dataframe of source and destination path pairs."""
-    return pd.read_csv(input_file, sep="\t")
-
-
 def main(input_file):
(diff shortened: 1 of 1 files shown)


ncalvanese1 (approved):
Looks good to me! Nice work!

bahill:
Looks good - just wondering, @ncalvanese1 & @KutnerBroadie why this isn't in the anvil_ingest_tools notebook? or is that the intended destination?

bahill (approved):

KutnerBroadie:
> Looks good - just wondering, @ncalvanese1 & @KutnerBroadie why this isn't in the anvil_ingest_tools notebook? or is that the intended destination?

That's the intended destination, I'm working on creating a subsection at the end for it.

PR #147, opened 2023-09-15
RARE-X: ignore ingest json files with 0 data to ingest into dataset

schaluva:
RARE-X receives json files for ingest from external source. 
If there is no data to be ingested, an empty json file is deposited in the bucket which breaks the ingest script.
Check for the size of the json to ensure that it is not empty before adding it to list of files to loop through to ingest into dataset.


ncalvanese1 (approved):

snovod (approved):
🔥

PR #148, opened 2023-09-19
[FE-62] Create bulk-set-requester-pays script

samanthahv:
https://broadworkbench.atlassian.net/browse/FE-62

This script allows users to set requester pays on/off for multiple workspaces in one command. 
Steps in the script:
1. checks input csv for correct headers and any missing values
2. grants the requester pays SA access to the workspace Google projects
3. authenticates as SA
4. loops through list of workspaces and sets requester pays based on toggle value provided
5. authenticates as user's firecloud account
6. removes SA from workspace Google projects

### DIFF
--- a/scripts/bulk_set_requester_pays.sh
+++ b/scripts/bulk_set_requester_pays.sh
@@ -6,7 +6,7 @@ if (( $# != 1 )); then
   echo 'The Terra bucket paths can be formatted as "gs://fc-XXXXX" or "fc-XXXXX"'
   echo "The toggle_type options are on|off for enabling and disabling requester pays"
   echo "NOTE: this script requires you to be authed as your firecloud.org admin account."
-  exit 0
+  exit 1
 fi
 
 
@@ -27,7 +27,7 @@ echo "Validating input file..."
 if [[ $(head -1 $CSVIN) != $'project,bucket,toggle_type\r' ]]; then
   echo "Missing or incorrect column headers"
   echo "inputfile.csv should contain the following column headers in order: project, bucket, toggle_type"
-  exit 0
+  exit 1
 fi
 
 # check for any missing values
@@ -47,7 +47,7 @@ done < $CSVIN
 if ( $missing ) ; then
   echo "ERROR: Missing values in input CSV file."
   echo "Input CSV file should contain the following columns and corresponding values: project, bucket, toggle_type"
-  exit 0
+  exit 1
 else
   echo "Valid input file."
 fi
--- a/scripts/bulk_set_requester_pays.sh
+++ b/scripts/bulk_set_requester_pays.sh
@@ -1,0 +1,131 @@
+#!/bin/bash
+
+if (( $# != 1 )); then
+  echo "Usage: $0 /path/to/inputfile.csv"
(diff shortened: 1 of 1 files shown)


scripts/bulk_set_requester_pays.sh
```diff
@@ -0,0 +1,129 @@
+#!/bin/bash
+
+if (( $# != 1 )); then
+  echo "Usage: $0 /path/to/inputfile.csv"
+  echo "inputfile.csv should contain the following column headers in order: project, bucket, toggle_type"
+  echo 'The Terra bucket paths can be formatted as "gs://fc-XXXXX" or "fc-XXXXX"'
+  echo "The toggle_type options are on|off for enabling and disabling requester pays"
+  echo "NOTE: this script requires you to be authed as your firecloud.org admin account."
+  exit 0
```
ahaessly, line 9: exit 0 indicates success. we should return exit 1 for failure conditions (here and below)

scripts/bulk_set_requester_pays.sh
```diff
@@ -0,0 +1,129 @@
...
+  echo "Valid input file."
+fi
+
+# 2. grant the "RequesterPaysToggler" role to the service account for all projects in the list
+# you need to be logged in with your firecloud account
+echo "Enabling permissions for ${SERVICE_ACCOUNT} to set Requester Pays"
+while IFS="," read -r PROJECT_ID; do
+  if gcloud projects add-iam-policy-binding $PROJECT_ID --member=$MEMBER --role=$ROLE --no-user-output-enabled; then
+    echo "Access granted to $PROJECT_ID"
+  fi
```
ahaessly, line 61: do we want an else block here to let the user know if granting access failed for any of the projects?

scripts/bulk_set_requester_pays.sh
```diff
@@ -0,0 +1,129 @@
...
+  while ! gsutil ${PROJECT_TO_BILL} requesterpays set ${TOGGLE_TYPE} ${BUCKET_PATH}
+    do
+      if (( $COUNTER > 5 )); then
+        echo "Maximum number of attempts exceeded for bucket $BUCKET_PATH - please check the error message and try again"
+        echo ""
+        break
+      fi
+      let COUNTER=COUNTER+1
+      # retry with a project to bill in case it failed with a 400 error
+      PROJECT_TO_BILL="-u broad-dsde-prod"
```
ahaessly, line 109: why do you set this again here? what if it was set to "" above? doesn't this change the meaning of the call when you retry?

ahaessly (approved):
a few comments

scripts/bulk_set_requester_pays.sh
samanthahv: thanks! fixed that in [0fc3154](https://github.com/broadinstitute/horsefish/pull/148/commits/0fc31549ebbab88e5051a9d4ed12997192e6ccea)

scripts/bulk_set_requester_pays.sh
samanthahv: added in [5a85d9f](https://github.com/broadinstitute/horsefish/pull/148/commits/5a85d9ff79700393211b479341fce6b1020098a5)

PR #165, opened 2024-02-20
remove hardcoding of sample_id

jpramosbarroso:


scripts/pgs/create_enterics_visualizations_html.py
```diff
@@ -715,11 +715,13 @@ def create_HTML(data, qc_metric_info, default_thresholds, entity_table_name, gro
     # Content-container for all plots and tables
     html_string += '<div id="content-container">'
 
+    sample_id_column = "entity:" + entity_table_name + "_id"
```
schaluva, line 718: What do you think about putting this definition in the main function - after its passed in as an argument. Line 886 or so where you can define the sample_id_column and then pass in that variable on lines 887 etc.

schaluva (approved):

